PR #9, opened 2020-03-19
Delete old files, Move Badge and Card to components/base

sausman:


components/base/Card.tsx
```diff
@@ -88,27 +88,29 @@ type Props = {
   description: string;
 };
 
-function Card(props: Props) {
+function Card({ description, badge, centerAlign, icon, link, topic }: Props) {
```
sausman, line 91: This is similar to what you had before I think. (I just realized you can do this with TypeScript, woot!)

stories/Badge.stories.tsx
```diff
@@ -1,6 +1,6 @@
 import * as React from "react";
 import { storiesOf } from "@storybook/react";
-import Badge from "../components/content/Badge";
+import Badge from "../components/base/Badge";
```
sausman, line 3: I know we talked about this haha, but I ended up changing these to "base" components since I don't foresee these being called directly within our content. I will update the README, but kind of what I'm thinking for the component types are:

- `base` - components that are references by either `content` or `layout` components.
- `content` - components that are directly referenced within our `.mdx` pages.
- `layout` - components that make up the layout

ShreyaThapa: `base` makes sense for these smaller (almost)individual components!

ShreyaThapa (approved):

ShreyaThapa (approved):

ShreyaThapa (approved):

PR #13, opened 2020-03-23
Add `Anchors{Context,Provider,Setter}` + `useAnchors`

sausman:
#### Recommended reading

https://reactjs.org/docs/context.html

#### Summary

Components for:

- Managing the Anchors context (`AnchorsProvider`)
- Setting anchors based on a page's content (`AnchorsSetter`)
- Allowing the current page's anchors to be accessed by other components (`useAnchors` hook)

## `AnchorsProvider` + `AnchorsContext`

Used in `_app.tsx` to provide `AnchorsContext` to the rest of the application.

```jsx
<AnchorsProvider>
  <Layout>
    <MyPage />
  </Layout>
</AnchorsProvider>
```

## `AnchorsSetter`

When a component is wrapped with `Anchors.Set`, it will:

- Add anchors (`id` props) to any headings (`h1..h6`) based on their inner text
- Report the anchors back to the context so they can be displayed elsewhere

```jsx
<AnchorsSetter>
  {mdx}
</AnchorsSetter>
```

**Note**: At the moment this component is specific to MDX in that it expects to receive `MDXCreateElement`s as its children.

## `useAnchors`

A hook for... using the `AnchorsContext`!

```javascript
type Anchor = {
  id: string;
  level: number;
  text: string;
};

const { anchors: Anchor[], setAnchors } = useAnchors();
```

Shorthand for `useContext(AnchorsContext)`


components/util/Anchors.tsx
```diff
@@ -0,0 +1,87 @@
...
+      return child.props.children.toString();
+    }
+    return acc + child.toString();
+  }, "");
+
+// Context
+
+const Context = createContext({
+  anchors: [],
+  setAnchors: (_anchors: string[]) => {}, // eslint-disable-line no-unused-vars
```
sausman: - [x] Include anchor `level` and `text`

components/util/__test__/Anchors.test.tsx
```diff
@@ -0,0 +1,50 @@
...
+}) {
+  return createElement(props.mdxType, {
+    children: props.children,
+    id: props.id,
+  });
+}
+
+test("Anchors", async () => {
+  let tree;
+  act(() => {
```
sausman, line 34: Since `AnchorsSetter` performs an update on the `AnchorsContext`, we need to wrap this in `act(() => ...)`. Else `DebugAnchors` would show an empty array for the anchors.

https://reactjs.org/docs/test-renderer.html#testrendereract

types/Anchor.d.ts
```diff
@@ -0,0 +1,5 @@
+type Anchor = {
```
sausman, line 1: I created a `types/` folder so we can share types across the app without having to import anything (which ESLint complains about since it thinks the variable is unused).

For example, we can use this in [#12].

PR #15, opened 2020-03-30
Added Button component and storybook addons

ShreyaThapa:
- Added Button Component
- Added Button story
- Added Button test
- Added Storybook addons
    - for centering component - @storybook/addon-centered/react
    - for adding dark background - @storybook/addon-backgrounds


stories/Button.stories.tsx
```diff
@@ -0,0 +1,29 @@
+import * as React from "react";
+import { storiesOf } from "@storybook/react";
+import centered from "@storybook/addon-centered/react";
+import Button from "../components/base/Button";
+
+storiesOf("Button", module)
```
sausman: Let's move this file to the `stories/base` directory and change this line to `storiesOf("base|Button", module) `

components/base/Button.tsx
```diff
@@ -0,0 +1,122 @@
...
+  return (
+    <ButtonStyle type="button" className={`${size} ${variant}`}>
+      <SpanStyle className={`${size}`}>{text}</SpanStyle>
+    </ButtonStyle>
+  );
+}
+
+const ButtonStyle = styled.button`
+  color: ${WHITE_PRIMARY};
+  font-family: "Poppins", sans-serif;
```
sausman: Let's import `POPPINS` from `typography.ts` here

stories/Button.stories.tsx
ShreyaThapa: Sounds good! I'll get those changes in.

components/base/Button.tsx
```diff
@@ -0,0 +1,122 @@
...
+  GREY_4,
+  GREY_9,
+  GREY_5,
+  PURPLE_DARK,
+} from "../colors";
+import { BOX_SHADOW_6 } from "../shadowDepths";
+
+type Props = {
+  text: string;
+  size: string;
```
sausman: For `size` and `variant` we could use [union types](https://www.typescriptlang.org/docs/handbook/advanced-types.html#union-types) to specify the accepted values. For example:

`size: "tiny" | "small" | "standard" | "large";`

ShreyaThapa: Thanks for catching that! 👍

ShreyaThapa: This [commit](https://github.com/Dwolla/dev-portal/pull/15/commits/7d47ff01d96b5204451ad2cd7f5ec6d82aa5b3c7) should fix that.

PR #16, opened 2020-03-31
Implement OnThisPage component

sausman:


stories/base/Card.stories.tsx
```diff
@@ -0,0 +1,74 @@
...
+
+storiesOf("base|Card", module)
+  .add("default", () => (
+    <Card
+      icon={guideIcon}
+      topic="This is the Topic"
+      description="An existing unverified bank can be verified with microdeposits. Send two small debits and have your Customer verify the amounts."
+    />
+  ))
+  .add("`link`", () => (
```
ShreyaThapa, line 14: Adding categories to Storybook looks neat! 👍 

I may be missing it, but is there a reason we added back-ticks to the sub-titles like in the following?
>.add("\`link\`", () => (

sausman: Ah yeah, I just added the back-ticks to denote that those are prop names.

ShreyaThapa: Ah, cool. That sounds good.

ShreyaThapa (approved):

PR #17, opened 2020-04-01
Update Footer

sausman:


.storybook/preview-head.html
```diff
@@ -0,0 +1,8 @@
+<link
```
sausman, line 1: Added these for consistency with what's used in Layout. Now our fonts work in Storybook!

ShreyaThapa: Nice!

ShreyaThapa (approved):

PR #18, opened 2020-04-06
Test onClick in FilterTabs

sausman:


components/base/FilterTabs.tsx
```diff
@@ -32,7 +32,7 @@ export default function Tabs({ tabs, filter, setFilter }: Props) {
       <TabsContainerStyle>
         <TabStyle
           className={`tab ${filter === null ? "is-active" : ""}`}
-          onClick={setFilter(null)}
+          onClick={() => setFilter(null)}
```
sausman, line 35: Put the call to `setFilter` within a function here. Previously it would call `setFilter(null)` on every render and pass the return value to `onClick` instead of a function.

components/base/FilterTabs.tsx
```diff
@@ -73,7 +73,7 @@ const TabsContainerStyle = styled.div`
   }
 `;
 
-const TabStyle = styled.div`
+export const TabStyle = styled.div`
```
sausman, line 76: exporting so we can use it in our test

components/base/__tests__/FilterTabs.test.tsx
```diff
@@ -1,18 +1,18 @@
 import renderer from "react-test-renderer";
-import ReactTestUtils from "react-dom/test-utils";
-import FilterTabs from "../FilterTabs";
+import FilterTabs, { TabStyle } from "../FilterTabs";
 
-test("Filter Tabs", () => {
+const TABS = ["GET BUILDING", "PREREQUISITES", "FEATURES"];
+
+test("FilterTabs", () => {
```
sausman, line 6: I renamed this to match the component name because some IDEs/editors will find matching string occurrences like this when/if you rename a component and update them for you.

components/base/__tests__/FilterTabs.test.tsx
```diff
@@ -1,18 +1,17 @@
...
-    )
-    .toJSON();
-  ReactTestUtils.Simulate.click(tree);
-  expect(tree).toMatchSnapshot();
+  const tree = renderer.create(
+    <FilterTabs tabs={TABS} filter={null} setFilter={setState} />
+  );
+  tree.root.findAllByType(TabStyle).map((t) => t.props.onClick());
+
+  expect(setState.mock.calls.map((c) => c)).toEqual(
```
sausman, line 13: `setState.mock.calls` is an array of arguments the mock function has been called with. In this case it should equal:

```javascript
[
  [null],
  ["GET BUILDING"],
  ["PREREQUISITES"],
  ["FEATURES"]
]
```

sausman:
@ShreyaThapa We should probably also test the Dropdown `onChange`, but did you notice:

```
console.warn node_modules/react-test-renderer/cjs/react-test-renderer.development.js:2256
      Warning: componentWillReceiveProps has been renamed, and is not recommended for use. See https://fb.me/react-unsafe-component-lifecycles for details.
      
      * Move data fetching code or side effects to componentDidUpdate.
      * If you're updating state whenever props change, refactor your code to use memoization techniques or move it to static getDerivedStateFromProps. Learn more at: https://fb.me/react-derived-state
      * Rename componentWillReceiveProps to UNSAFE_componentWillReceiveProps to suppress this warning in non-strict mode. In React 17.x, only the UNSAFE_ name will work. To rename all deprecated lifecycles to their new names, you can run `npx react-codemod rename-unsafe-lifecycles` in your project source folder.
      
      Please update the following components: Dropdown
```

I found https://github.com/fraserxu/react-dropdown/pull/170 but it doesn't look like there's been much activity on it recently.

ShreyaThapa (approved):

PR #20, opened 2020-04-06
Update TopBar component

sausman:
There's some other small stuff in here too:

- [x] Remove css widths from Button styles (?)
- [x] Update Footer styling
- [x] Don't display OnThisPage if no anchors
- [x] Add /changelog page
- [x] Update TopBar component

### DIFF
--- a/components/base/Button.tsx
+++ b/components/base/Button.tsx
@@ -35,7 +35,7 @@ const ButtonStyle = styled.button`
   color: ${WHITE_PRIMARY};
   font-family: ${POPPINS};
   font-style: thin;
-  text-align: center;
+  padding: 0 15px;
   border-radius: 5px;
   cursor: pointer;
   :focus {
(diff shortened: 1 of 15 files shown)


components/base/Button.tsx
```diff
@@ -35,27 +35,23 @@ const ButtonStyle = styled.button`
...
-  text-align: center;
+  padding: 0 15px;
   border-radius: 5px;
   cursor: pointer;
   :focus {
     outline: 0;
   }
   &.tiny {
     height: 26px;
-    width: 72px;
```
sausman, line 46: @ShreyaThapa I ended up taking out the `width`s and adding some padding instead to get it to match the TopBar mockups, but do you know if these are being used elsewhere where we want the `width`s?

ShreyaThapa (approved):

components/base/Button.tsx
ShreyaThapa: Removing the Width should be fine. If we come across anything that needs the width, we can revisit the component! 👍

PR #23, opened 2020-04-08
WEB-614: Typography component

ShreyaThapa:
- [x] Added Package - @storybook/addon-notes for adding notes to Storybook

- [x] Typography test file
- [x] Typography Story file

Typography component 
- [x] Headings, Paragraphs, Lists
- [x] Inline code block, Links, Block quote
-----------------------------------------------------------------------
What do you think about the Typography.tsx file structure? Do you think that will work? Feel free to let me know of any better ways to write this file.

I have yet to add support for the following - 
- Inline code blocks
- block quotes
- links


components/base/Typography.tsx
```diff
@@ -0,0 +1,151 @@
...
+  WHITE_PRIMARY,
+  GREY_5,
+  GREY_4,
+} from "../colors";
+
+// Typography function
+function Typography({ text, styleVariant, tag, isDark }: TypographyProps) {
+  return (
+    <TypographyStyles
+      as={`${tag}`}
```
ShreyaThapa: > as={\`${tag}\`}

@sausman, this is what I'd come up with for dynamic tags in styled-components, but it seems to be failing checks. Would you be able to take a look at this when you get a chance?

sausman: Hmm yeah that's bizarre, something with TypeScript. I tried updating dependencies and that didn't fix it.

components/base/Typography.tsx
```diff
@@ -0,0 +1,151 @@
...
+    >
+      {text}
+    </TypographyStyles>
+  );
+}
+
+// Typography prop types
+type TypographyProps = {
+  text: string;
+  styleVariant:
```
sausman: I wonder if we should use [composition](https://emotion.sh/docs/composition) or something to split these style variants into their own components that share some common CSS.

e.g.

```typescript
const headingStyles = css`/* heading styles */`
export const H1 = ({ children }) => <h1 css=[headingStyles, css`/* h1 styles */`]>{children}</h1>
```

components/base/Typography.tsx
```diff
@@ -0,0 +1,151 @@
...
+      className={classnames(`${styleVariant}`, { dark: isDark })}
+    >
+      {text}
+    </TypographyStyles>
+  );
+}
+
+// Typography prop types
+type TypographyProps = {
+  text: string;
```
sausman: Let's change this from `text` to `children` to be consistent with the built-in components (`h1`, `h2`, `p` etc.).

That way we can add these components as-is to `MDX_COMPONENTS` since MDX will pass the text as the `children` prop.

components/base/Typography.tsx
ShreyaThapa: I like that idea! I'll get those updated 👨‍⚖️

components/base/Typography.tsx
```diff
@@ -0,0 +1,211 @@
...
+  padding: 2px 5px;
+  &.dark {
+    background-color: ${PURPLE_DARK};
+    color: ${GREY_4};
+  }
+`;
+
+const StyledLink = styled.a`
+  color: ${ORANGE_PRIMARY};
+  &:hover {
```
sausman, line 109: I think we can remove these `:hover` styles since `<a>` is underlined and has `cursor: pointer` by default

components/base/Typography.tsx
```diff
@@ -0,0 +1,211 @@
...
+  </blockquote>
+);
+
+export const InlineCode = ({ children, isDark }: Props) => (
+  <StyledInlineCode className={classnames({ dark: isDark })}>
+    {children}
+  </StyledInlineCode>
+);
+
+export const Link = ({ children }: Props) => (
```
sausman: Is this supposed to handle `isDark`?

ShreyaThapa: Perhaps not, because the styles for both Dark and Light is the same for links, which is `color: ${ORANGE_PRIMARY}` Unless I'm missing something 🤔

sausman: Ah ok cool, we can probably just do `export const Link = StyledLink` here then if you want.

ShreyaThapa: Actually, thinking of doing this: 
```suggestion
export const Link = ({ text, href }: LinkProps) => (
  <StyledLink href={href}>{text}</StyledLink>
);```

ShreyaThapa: What do you think? This [commit](https://github.com/Dwolla/dev-portal/pull/23/commits/e0aa5f3e35215e7ff93f6a7c624727ac3b155496) contains those changes.

sausman: Hmm, I'd probably lean towards keeping the props the same as the built-in `<a>` since all we're really adding are styles. Why change things up/write more code if we don't have to?

This works too tho, we can always change it if we need to!

sausman (approved):

PR #24, opened 2020-04-08
WEB-599: Adding simple password auth to protect content

Finholt:


components/AuthPage.tsx
```diff
@@ -0,0 +1,81 @@
...
+  font-size: 14px;
+  font-family: ${POPPINS};
+`;
+
+const ErrorStyled = styled.div`
+  margin-top: 10px;
+  font-size: 14px;
+  font-family: ${POPPINS};
+  color: #e9423b;
+`;
```
Finholt, line 79: If this were something that we planned to keep in production at launch, I would break some of these into more reusable components, fine tune them a bit more, etc., but this seemed good enough for a temporary auth page

components/AuthPage.tsx
```diff
@@ -0,0 +1,81 @@
...
+};
+
+const AuthPage = ({ setAuthenticated }: Props) => {
+  const [password, setPassword] = useState("");
+  const [passwordError, setPasswordError] = useState(false);
+
+  const authenticate = (e) => {
+    e.preventDefault();
+
+    if (password === "backtowork") {
```
Finholt, line 18: This is the super secure password, btw

next.config.js
```diff
@@ -1,17 +1,24 @@
...
+      // rehypePlugins: [],
+      // extendFrontMatter: {
+      //   process: (mdxContent, frontMatter) => {},
+      //   phase: 'prebuild|loader|both'
+      // }
+    })({
+      target: "serverless",
+      env: {
+        isDev: phase === PHASE_DEVELOPMENT_SERVER,
+      },
```
Finholt, line 22: Setting this so we don't need to enter the password all the time when we're working locally

components/AuthPage.tsx
sddaniels: noice

sausman (approved):

PR #28, opened 2020-04-15
Web-629: Card Grid component

ShreyaThapa:
Added Card Grid component


components/base/CardGrid.tsx
```diff
@@ -0,0 +1,40 @@
...
+import {
+  maxWidth,
+  BREAKPOINT_DESKTOP,
+  BREAKPOINT_IPAD,
+  BREAKPOINT_MOBILE,
+} from "../breakpoints";
+
+const StyledGrid = styled.div`
+  display: flex;
+  flex-wrap: wrap;
```
Finholt, line 11: Can you add some negative margins on the left, right, and bottom of the grid in order to cancel out the padding added by CardWrappers along the edge?

ShreyaThapa: Good catch, Conner! Thanks!

This [commit](https://github.com/Dwolla/dev-portal/pull/28/commits/eab9341ea373dac6ddaf2416abc1d5dd5c6ecf30) should address that.

Finholt (approved):

PR #31, opened 2020-04-16
WEB-608: Adding Selector Dropdown Component

Finholt:


components/colors.ts
```diff
@@ -8,6 +8,7 @@ export const GREY_6 = "#52627b";
 export const GREY_7 = "#3e4e67";
 export const GREY_8 = "#354153";
 export const GREY_9 = "#2e3745";
+export const GREY_10 = "#cfcff5";
```
Finholt, line 11: I know it's following the designs, but we might be approaching the point where there are too many grays. That's a conversation to be had with the entire group though, so I'm just rolling with the colors the designs used for now

package.json
```diff
@@ -60,6 +60,7 @@
     "react-dom": "^16.12.0",
     "react-dropdown": "^1.7.0",
     "react-scroll": "^1.7.16",
+    "react-select": "3.1.0",
```
Finholt, line 63: Using this library so we can style the dropdown menu without having to directly worry about functionality, accessibility, etc. Looks to be quite lightweight: https://bundlephobia.com/result?p=react-select@3.1.0

public/images/double-arrows-select-icon.tsx
```diff
@@ -0,0 +1,11 @@
+import React from "react";
+
+const DoubleArrowsSelectIcon = () => (
+  <svg xmlns="http://www.w3.org/2000/svg" viewBox="0 0 7 12" width="7px">
+    <g fill="#CFCFF5" fillRule="evenodd">
```
Finholt, line 5: Because I needed to update this fill with CSS, I'm essentially importing the SVG as a component so that I can use it inline rather than referenced through an img tag. We could do this without manually converting the SVG pretty easily in our create-react-app. Is there a better way to do this in Next?

components/base/select/Select.tsx
```diff
@@ -0,0 +1,80 @@
...
+import DoubleArrowSelectIcon from "../../../public/images/double-arrows-select-icon";
+import CheckMarkIcon from "../../../public/images/check-mark-icon";
+import {
+  StyledDefaultDropdownIcon,
+  StyledOptionWrap,
+  StyledSelectedIconWrap,
+  defaultCustomStyles,
+  codeCustomStyles,
+  addAutoWidthStyles,
+} from "./Select.styles";
```
Finholt, line 13: Because all of the styles made this file quite long and hard to read, I ended up splitting them out. That said, I know our team approach to this was still an open question, so I'm happy to reconsider that decision!

public/images/check-mark-icon.tsx
```diff
@@ -0,0 +1,21 @@
+import React from "react";
+
+const DoubleArrowsSelectIcon = () => (
```
sausman: Is this meant to be `CheckMarkIcon`?

public/images/double-arrows-select-icon.tsx
sausman: Hmm yeah, how does it work in CRA? If it's with a Webpack loader we could probably add that loader to our `next.config.js` similar to what `withImages` does.

We may also be able to use the [`css` prop](https://emotion.sh/docs/css-prop#string-styles) here or pass the `fill` in as a prop?

---

Somewhat related: I think we started by putting stuff in `/public` since Next.js serves that publicly (allowing ``, but it doesn't work as well with Storybook since it does everything with Webpack. Probably a larger discussion for the team, but maybe we start putting these in `components/atoms` and make a ticket to move what we can from `/public`.

public/images/check-mark-icon.tsx
Finholt: Yep! Thank you for catching my bad copy-paste! Should be fixed by the latest commit

public/images/double-arrows-select-icon.tsx
Finholt: CRA lets you import them as components by changing the import statement from `import logo from './logo.svg';` to `import { ReactComponent as Logo } from './logo.svg';` (per [here](https://create-react-app.dev/docs/adding-images-fonts-and-files/#adding-svgs)). I'm not sure how they do it behind the scenes

I looked at maybe doing something like [this](https://github.com/airbnb/babel-plugin-inline-react-svg), but that would change all of our SVGs to component imports and prevent us from using them in img tags (which we're already doing in a number of places)

components/base/select/Select.tsx
ShreyaThapa: Yeah, I think I like it separated too. Let's keep it this way for now, and revisit the topic in our meeting next week.

ShreyaThapa (approved):
That dropdown looks hella nice!

components/base/select/Select.tsx
sausman: Yeah I like that you split them out here. I’m tempted to say “do what feels right” in situations like this cuz I don’t universally prefer one approach over the other?

When things are in one file I like that I can glance/swipe up to see the styles w/o opening another file, but at a certain point it’s easier to have 2 files open to avoid a bunch of scrolling/finding your place.

public/images/double-arrows-select-icon.tsx
sausman: Ah yeah those look interesting. I wouldn't be opposed to switching the existing SVGs to use something different so maybe we do that in another ticket after some more research?

sausman (approved):

public/images/double-arrows-select-icon.tsx
Finholt: Sounds like a plan! I created https://dwolla.atlassian.net/browse/WEB-670 to track it

PR #32, opened 2020-04-20
Added Table component

ShreyaThapa:


components/base/Table.tsx
```diff
@@ -0,0 +1,67 @@
+import React from "react";
+import styled from "@emotion/styled";
+import { ROBOTO } from "../typography";
+import { HEADLINE_TEXT, PARAGRAPH_TEXT, GREY_1, GREY_2 } from "../colors";
+
+// Styles for Table
+const StyledTable = styled.table`
```
Finholt: So right now this is rendering:
```
<table>
  <table>
    <tr>
      ...
```

It feels like this should either be a styled div implemented like it is right now, or it should be a styled table where the `<table>` is omitted from the children being passed as props

components/base/Table.tsx
```diff
@@ -0,0 +1,67 @@
...
+  table {
+    /* Separate borders to selectively apply border radius */
+    border-collapse: separate;
+    border-spacing: 0;
+    text-align: left;
+    font-family: ${ROBOTO};
+    overflow-x: scroll;
```
Finholt: Should we include some styles here that **A)** allow the table's width to expand to the width of its parent, and **B)** set some kind of minimum width (here or at the `td` level to prevent the cells from scrunching up on narrower screens before the overflow kicks in?

Finholt (commented):
I have a couple of small things that we might want to figure out, but overall, this looks pretty great!

components/base/Table.tsx
ShreyaThapa: Whoops! Good catch, Connor! 

This is a remnant from when I moved away from the component handling the creating the table itself. I'd say let's change it to a styled div.

components/base/Table.tsx
ShreyaThapa: That's a good idea! I'll take care of those 👍

PR #33, opened 2020-04-20
WEB-626: Setting up Context Provider for Language Selectors

Finholt:


components/layout/Layout.tsx
```diff
@@ -70,6 +70,7 @@ const TopBarWrapper = styled.div`
   grid-area: top-bar;
   position: sticky;
   top: 0;
+  z-index: 10;
```
Finholt, line 73: On the current homepage, the footer covered up the dropdown. Probably wouldn't be an issue with real content, but better safe than sorry

components/util/Contexts.tsx
```diff
@@ -0,0 +1,10 @@
+import React from "react";
+
+// eslint-disable-next-line import/prefer-default-export
+export const LanguageContext = React.createContext({
+  selectedLanguage: { value: "", label: "" },
+  setSelectedLanguage: (language) => {
+    return language;
+  },
+  languageOptions: [],
```
Finholt, line 9: Default data overwritten by `LanguageContext.Provider` in `_app.tsx`

components/layout/TopBar.tsx
```diff
@@ -95,6 +102,9 @@ export type TopBarProps = {
 };
 
 export default function TopBar({ button, links }: TopBarProps) {
+  const { selectedLanguage, setSelectedLanguage, languageOptions } = useContext(
+    LanguageContext
+  );
```
Finholt, line 107: Accessing context data setup by `LanguageContext.Provider` in `_app.tsx`

pages/_app.tsx
```diff
@@ -95,15 +105,23 @@ const AppWithHooks = ({ router, Component, pageProps }: any) => {
...
-          pages={Pages.all()}
-          sideNavLinks={SIDE_NAV_LINKS}
-          topBarProps={TOP_BAR_PROPS}
-          apiStatus={apiStatus}
+        <LanguageContext.Provider
+          value={{
+            selectedLanguage,
+            setSelectedLanguage,
+            languageOptions: LANGUAGE_OPTIONS,
+          }}
```
Finholt, line 118: Wrapping app in a context provider so all components contained within should have access to this data without having to worry about Prop drilling

Finholt (commented):
Since we don't think there will be too much global data for this project, something like redux seemed like overkill to keep track of the selected language. As such, I opted to give `Context` and the `useContext` hook ([for reference](https://reactjs.org/docs/hooks-reference.html#usecontext)) a try

It took a minute to figure it out, but I really like it for a situation like this! Let me know if something isn't making sense and I can try to explain!

ShreyaThapa:
I think using Context works best in this case. This looks good, Connor!
𝔸𝕡𝕡𝕣𝕠𝕧𝕖𝕕

pages/_app.tsx
ShreyaThapa: Nice!

ShreyaThapa (approved):

stories/layout/TopBar.stories.tsx
```diff
@@ -1,8 +1,43 @@
...
+};
+
+const LangaugeContextAccessor = () => {
+  const { selectedLanguage }: LanguageContextType = useContext(LanguageContext);
+
+  return <h2>Selected Language: {selectedLanguage.label}</h2>;
+};
 
 storiesOf("layout|TopBar", module)
+  .addDecorator(LanguageContextDecorator)
```
sausman, line 40: :nice: that's a cool pattern

sausman (approved):

PR #34, opened 2020-04-21
Add MDXTypographyWrapper

sausman:
I think I got everything that won't be addressed by some other tickets we're working on (e.g. `<table>`s). I tried adding some snapshot tests and a story for this, but was running into some issues importing an `.mdx` file cuz I don't think we have it set up to do that with TypeScript etc. I ended up leaving that stuff out since we have some stories for the individual elements, but I can always revisit either in this ticket or another if we think those are important to have.

image


components/base/Typography.tsx
```diff
@@ -207,11 +229,74 @@ export const BlockQuote = ({ children, isDark }: Props) => (
...
+        padding: 20px 40px;
+      }
+
+      > h1,
+      h2,
+      h3,
+      h4,
+      h5,
+      h6 {
+        ${headingStyles.global.styles}
```
sausman, line 258: I'm not entirely sure why, but I had to add `.styles` to get these to work in here. I think it has something to do with styling a non-root element? The only downside I'm aware of is it won't share the same CSS class with other components using these styles which isn't a huge deal, and I didn't want to hold this up since it works.

ShreyaThapa:
It's coming together nicely! Looks good, Stephen. 

> I ended up leaving that stuff out since we have some stories for the individual elements, but I can always revisit either in this ticket or another if we think those are important to have.

I think it'd be alright leaving the tests out for this one.

ShreyaThapa (approved):

ShreyaThapa:
𝔸𝕡𝕡𝕣𝕠𝕧𝕖𝕕

Finholt (approved):

PR #38, opened 2020-04-22
Added Pagination component

ShreyaThapa:


components/base/Pagination.tsx
```diff
@@ -0,0 +1,75 @@
...
+  variant: "next" | "previous";
+  variantText: string;
+  href: string;
+  children: string;
+};
+
+// Pagination component
+function Pagination({ variant, variantText, href, children }: Props) {
+  return (
+    <StyledAnchor href={href}>
```
Finholt: Can we use a Nextjs `Link` component instead of a generic anchor here to make use of Nextjs's router? I looked at the `TopBarLink` component as a reference for how to make that work with our styled components

stories/base/Pagination.stories.tsx
```diff
@@ -0,0 +1,54 @@
...
+    a {
+      margin: 10px 0;
+    }
+  }
+`;
+
+storiesOf("base|Pagination", module)
+  .add("Next", () => (
+    <PaginationWrapper className="center">
+      <Paginaton variant="next" variantText="NEXT" href="#">
```
Finholt, line 33: This is probably a nitpick, but can we just use `text-transform: uppercase` on the variant text? I feel like that leaves less room for inconsistencies/typos while maintaining this down the road

components/base/Pagination.tsx
```diff
@@ -0,0 +1,75 @@
...
+      <Container>
+        {variant === "next" ? (
+          <ArrowDiv>
+            {variantText}
+            <img src={arrowRight} alt="Next" />
+          </ArrowDiv>
+        ) : (
+          <ArrowDiv>
+            <img src={arrowLeft} alt="Previous" />
+            {variantText}
```
Finholt: Looks like the designs have the text right-aligned on the previous variant

ShreyaThapa: Totally missed that! Thanks for checking that Connor! :)

stories/base/Pagination.stories.tsx
ShreyaThapa: That's good point! 👍

components/base/Pagination.tsx
ShreyaThapa: Ah, sounds good! I'll get those updated! 👍

ShreyaThapa: Should be address by this [commit](https://github.com/Dwolla/dev-portal/pull/38/commits/2cf01f7d6d3eded6cbcd2dbe20faa810c278d2fa)!

stories/base/Pagination.stories.tsx
ShreyaThapa: Should be address by this [commit](https://github.com/Dwolla/dev-portal/pull/38/commits/2cf01f7d6d3eded6cbcd2dbe20faa810c278d2fa)!

components/base/Pagination.tsx
ShreyaThapa: Should be address by this [commit](https://github.com/Dwolla/dev-portal/pull/38/commits/2cf01f7d6d3eded6cbcd2dbe20faa810c278d2fa)!

Finholt (approved):
:chefkiss: Approved!

PR #39, opened 2020-04-23
Added Alert Bar component

ShreyaThapa:


components/base/AlertBar.tsx
```diff
@@ -0,0 +1,60 @@
...
+import {
+  GREY_1,
+  GREY_2,
+  API_STATUS_BAR_YELLOW,
+  PARAGRAPH_TEXT,
+} from "../colors";
+import { ROBOTO } from "../typography";
+
+const StyledAlertBar = styled.div`
+  display: flex;
```
Finholt, line 14: What are your thoughts on adding `align-items: center;` here in order to vertically center the icon if the text wraps to more than one line? I'm going back and forth

ShreyaThapa: Yeah, I was going back and forth on that too! 

I tried going off of this [page](https://projects.invisionapp.com/d/main#/console/19172953/400928381/preview) in the mock-ups that has the icon positioned at the top when the text wraps.  Thought it looked better that way too. Let's ask the team about it. 👍

ShreyaThapa: Looks like the verdict is to keep it positioned at the top. 🔝

Finholt: Sounds good! Thanks for talking through that with me/the team 💯

Finholt (approved):

components/base/AlertBar.tsx
ShreyaThapa: 😎👍

PR #40, opened 2020-04-24
WEB-632: Filter functionality for index pages

Finholt:


components/base/CardGrid.tsx
```diff
@@ -14,7 +9,7 @@ const StyledGrid = styled.div`
 const CardWrapper = styled.div`
   width: 25%;
   padding: 15px;
-  @media (${maxWidth(BREAKPOINT_DESKTOP)}) {
+  @media (${maxWidth(1600)}) {
```
Finholt, line 12: With the side nav taking up a quarter of available screen real estate, things go too scrunched with four columns on the regular desktop breakpoint

components/base/FilterTabs.tsx
```diff
@@ -47,20 +48,30 @@ const SelectWrapper = styled.div`
...
 };
 
 export default function FilterTabs({ tabs, filter, setFilter }: Props) {
+  const formattedTabs = tabs.map((tab) =>
+    typeof tab === "string" ? { value: slugify(tab), label: tab } : tab
+  );
+  const formattedFilter =
+    typeof filter === "string" || filter === null
+      ? formattedTabs.find((tab) => tab.value === filter)
+      : filter;
```
Finholt, line 65: Wanted to have the ability for tabs and the selected filter to be passed in as strings instead of the value/label object that react-select looks for

components/layout/Layout.tsx
```diff
@@ -56,42 +57,42 @@ const MainArea = styled.div`
...
 
-  display: grid;
-  grid-template-columns: 70% 30%;
-  grid-template-rows: auto;
-  grid-template-areas:
-    "top-bar top-bar"
-    "content sidebar"
-    "footer footer";
+const ContentArea = styled.div`
+  display: flex;
```
Finholt: Changing this up so that the main ContentWrapper can expand when the OnThisPageWrapper is empty/hidden

pages/docs/guides/index.mdx
```diff
@@ -4,73 +4,71 @@ title: "Guides"
...
-</nav>
+import IndexPageGrid from "../../../components/partial/IndexPageGrid";
+import guidesIcon from "../../../public/images/guides-icon-large.svg";
+
+export const bannerProps = {
+  icon: guidesIcon,
+  topic: "Guides",
+  description: "Step-by-step instructions to get you set up in the Dwolla API",
+};
+export const guideCards = [
```
Finholt, line 15: We might eventually want this to be replaced with something populated directly from each Guide's front-matter, but since how all of that is going to look is still in a bit of flux right now, I propose we take another look at that as a part of [WEB-659](https://dwolla.atlassian.net/browse/WEB-659)

components/partial/IndexPageGrid.tsx
```diff
@@ -0,0 +1,89 @@
...
+  const setFilter = (selectedFilter) => {
+    if (selectedFilter.value) {
+      router.push({
+        pathname: router.pathname,
+        query: { filter: selectedFilter.value },
+      });
+    } else {
+      router.push(router.pathname);
+    }
+  };
```
Finholt, line 55: Using url query params (rather than state) to track the selected filter. This has the benefit of being able to link directly to filtered results

components/base/CardGrid.tsx
ShreyaThapa: Good point! Thanks for taking care of these as well as you go along! 🙌

pages/docs/guides/index.mdx
ShreyaThapa: Agreed. 👍 We may be able to do that. Sounds good!

components/layout/Layout.tsx
```diff
@@ -56,42 +57,42 @@ const MainArea = styled.div`
   @media (min-width: 980px) {
     margin-left: 25%;
```
ShreyaThapa, line 63: Could we also add a `margin-right: 30px` here so that the content doesn't extend towards the right? 
In a smaller screen, the content/banner would take up the whole screen which is fine. But in a bigger screen, I think it looks better with a little gap on the right like in the mock-ups.

Finholt: For sure! I was going back and forth; should be fixed [here](https://github.com/Dwolla/dev-portal/pull/40/commits/e452d15de54766a13e5e5e80cf5bf6e927ba2236) (Went with 40px so it lines up with the button in the top nav btw)

ShreyaThapa: Sweet! Yeah, good call! Does look better when it's lined up with the button. 👍

ShreyaThapa (approved):

ShreyaThapa:
𝔸𝕡𝕡𝕣𝕠𝕧𝕖𝕕

PR #41, opened 2020-04-24
Added Collapsibile component for FAQ

ShreyaThapa:
Using react-collapsible for the accordion - https://bundlephobia.com/result?p=react-collapsible@2.7.0

### DIFF
--- a/components/base/Collapsible.tsx
+++ b/components/base/Collapsible.tsx
@@ -12,7 +12,6 @@ import { BOX_SHADOW_7 } from "../shadowDepths";
 const CollapsibleWrapper = styled.div`
   /* Main container element for react-collapsible https://www.npmjs.com/package/react-collapsible#collapsible */
   .Collapsible {
-    padding: 26px 30px;
     box-sizing: border-box;
     width: 100%;
     border: 1px solid ${GREY_2};
@@ -26,10 +25,10 @@ const CollapsibleWrapper = styled.div`
       box-shadow: ${BOX_SHADOW_7};
     }
   }
+
   /* Content element for react-collapsible https://www.npmjs.com/package/react-collapsible#collapsible__contentinner */
   .Collapsible__contentInner {
-    margin-top: 24px;
-    padding-right: 40px;
+    margin: 0px 70px 25px 30px;
     > * {
       margin: unset;
     }
@@ -37,6 +36,7 @@ const CollapsibleWrapper = styled.div`
 `;
 
 const StyledTrigger = styled.div`
+  padding: 25px 30px;
   display: flex;
   justify-content: space-between;
   align-items: start;
@@ -55,7 +55,7 @@ function Collapsible({ triggerText, children }: Props) {
   const closedTrigger = (
     <StyledTrigger>
       {triggerText}
-      <img css={ImageStyles} src={plusIcon} alt="Collapsible closed icon" />
+      <img css={ImageStyles} src={plusIcon} alt="Open collapsible content" />
     </StyledTrigger>
   );
 
@@ -63,7 +63,7 @@ function Collapsible({ triggerText, children }: Props) {
   const openTrigger = (
     <StyledTrigger>
       {triggerText}
-      <img css={ImageStyles} src={minusIcon} alt="Collapsible open icon" />
+      <img css={ImageStyles} src={minusIcon} alt="Close collapsible content" />
     </StyledTrigger>
   );
 
--- a/components/base/Collapsible.tsx
+++ b/components/base/Collapsible.tsx
@@ -1,0 +1,85 @@
+/** @jsx jsx */
+import ReactCollapsible from "react-collapsible";
+import styled from "@emotion/styled";
+import { css, jsx } from "@emotion/core";
+import plusIcon from "../../public/images/plus-circle-icon.svg";
+import minusIcon from "../../public/images/minus-circle-icon.svg";
+import { GREY_2, WHITE_PRIMARY } from "../colors";
+import { POPPINS } from "../typography";
+import { BOX_SHADOW_7 } from "../shadowDepths";
+
+// Styles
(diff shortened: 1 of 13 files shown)


components/base/Collapsible.tsx
```diff
@@ -0,0 +1,85 @@
...
+import { css, jsx } from "@emotion/core";
+import plusIcon from "../../public/images/plus-circle-icon.svg";
+import minusIcon from "../../public/images/minus-circle-icon.svg";
+import { GREY_2, WHITE_PRIMARY } from "../colors";
+import { POPPINS } from "../typography";
+import { BOX_SHADOW_7 } from "../shadowDepths";
+
+// Styles
+const CollapsibleWrapper = styled.div`
+  /* Main container element for react-collapsible https://www.npmjs.com/package/react-collapsible#collapsible */
```
ShreyaThapa, line 13: Overriding **react-collapsible**'s built in classes `.Collapsible` and `.Collapsible__contentInner`

components/base/Typography.tsx
```diff
@@ -113,7 +113,7 @@ const listStyles = css`
   ${baseTextStyles}
   font-size: 16px;
   letter-spacing: 0;
-  line-height: 28px;
+  line-height: 36px;
```
ShreyaThapa: Increased line-height of Lists to make it look more similar to the mock ups.

Finholt: I agree that this makes the space between bullet points look better, but if an item wraps to multiple lines, it also adds the extra space there, which doesn't look quite right to me (I also found an example of this case in the designs [here](https://projects.invisionapp.com/d/main#/console/19172953/400098958/preview))

Could we change the line-height back and then add the extra space between the bullets by applying a bottom margin to the `<li>`s instead?

components/base/Collapsible.tsx
```diff
@@ -0,0 +1,85 @@
...
+    padding: 26px 30px;
+    box-sizing: border-box;
+    width: 100%;
+    border: 1px solid ${GREY_2};
+    border-radius: 5px;
+    background-color: ${WHITE_PRIMARY};
+    font-family: ${POPPINS};
+    font-size: 18px;
+    line-height: 27px;
+    cursor: pointer;
```
Finholt: I noticed that the cursor changes if you hover anywhere on the component, but only toggles if you click on the 'question' or the icon. Can we make it so that either the cursor only changes when hovering over that part of the component or clicking anywhere on the card triggers the toggle?

I think being able to click anywhere would be a better solution and more consistent with our other card-like components, but that might cause problems if the answer has a link in it (?), so I'd be fine with either!

components/base/Collapsible.tsx
```diff
@@ -0,0 +1,85 @@
...
+      {triggerText}
+      <img css={ImageStyles} src={plusIcon} alt="Collapsible closed icon" />
+    </StyledTrigger>
+  );
+
+  // Main content when collapsible is open
+  const openTrigger = (
+    <StyledTrigger>
+      {triggerText}
+      <img css={ImageStyles} src={minusIcon} alt="Collapsible open icon" />
```
Finholt: The alt tags here might be kind of confusing. Could we change them to something along the lines of "Open collapsible content" and "Close collapsible content", respectively? I generally find that alt text for icons, if it's even needed at all, is more useful when it describes what interacting with the icon does rather than what the icon itself is

package.json
```diff
@@ -57,6 +57,7 @@
     "next-mdx-enhanced": "^2.4.0",
     "prettier": "^1.19.1",
     "react": "^16.12.0",
+    "react-collapsible": "^2.7.0",
```
Finholt, line 60: Good choice of a library! Seems like it was pretty easy to implement and looks like it's working really well 💯

components/base/Typography.tsx
ShreyaThapa: That's a great point! Didn't even think about that 🤦

components/base/Collapsible.tsx
ShreyaThapa: For sure! That's a good point. I was just using that tag to literally describe what the icon is lol. But I like the idea of describing what the icon does. I'll get those updated!

components/base/Collapsible.tsx
ShreyaThapa: I like the the latter too: being able to click anywhere in the card. I remember trying other ways to make the whole card clickable but failed. 🤔  I'll tinker around with that, and also try it with adding links to the answers as well! Thanks for the review, Connor!

ShreyaThapa: This [commit](https://github.com/Dwolla/dev-portal/pull/41/commits/d0d3001d409ff380c5db788e1f385048dcd05363) should take care of that! 
Instead of overriding  the padding style of the container, I ended up adding that padding to the trigger prop that's getting passed to the package, which seems to be working to let us click anywhere on the closed card!

components/base/Typography.tsx
ShreyaThapa: Address [here](https://github.com/Dwolla/dev-portal/pull/41/commits/d0d3001d409ff380c5db788e1f385048dcd05363). :)

components/base/Collapsible.tsx
Finholt: Ahhh, nice fix for that! One last small thing, can you move `cursor: pointer` from `.Collapsible` to the `StyledTrigger` styles? It still looks like clicking on the bottom part of the open card will do something when it doesn't

ShreyaThapa: Ah, good catch! Missed that! 😅 This [commit](https://github.com/Dwolla/dev-portal/pull/41/commits/86b56e7f9ddd7d8efe5e265693bd20e5c0accac3) should fix that! :)

Finholt (approved):

PR #42, opened 2020-04-27
Added Footer CTA | Updated Button paddings

ShreyaThapa:
- Added Footer CTA component
- Updated Buttons to add paddings based on Button size

### DIFF
--- a/components/base/FooterCTA.tsx
+++ b/components/base/FooterCTA.tsx
@@ -18,15 +18,28 @@ const Container = styled.div`
     padding: 10px;
     flex-direction: column;
   }
+`;
+
+const StyledContent = styled.div`
+  padding-right: 20px;
+  > * {
+    margin: unset;
+    margin-bottom: 13px;
+  }
+  @media (${maxWidth(BREAKPOINT_MOBILE)}) {
+    padding-right: unset;
+    text-align: center;
+  }
   /* Text sizes decrease in smaller screens */
-  > div > h2 {
-    @media (${maxWidth(BREAKPOINT_IPAD)}) {
+  @media (${maxWidth(BREAKPOINT_IPAD)}) {
+    > * {
+      margin-bottom: 7px;
+    }
+    > h2 {
       font-size: 18px;
       line-height: 22px;
     }
-  }
-  > div > span {
-    @media (${maxWidth(BREAKPOINT_IPAD)}) {
+    > p {
       font-size: 16px;
       line-height: 20px;
     }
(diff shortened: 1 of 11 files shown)


components/base/FooterCTA.tsx
```diff
@@ -0,0 +1,54 @@
...
+// Props
+type ButtonProps = {
+  text: string;
+};
+type Props = { topic: string; description: string; button?: ButtonProps };
+
+function FooterCTA({ topic, description, button }: Props) {
+  return (
+    <Container>
+      <div>
```
Finholt: Could we style this div to add some padding-right to iPad and up breakpoints and some padding-bottom to mobile breakpoints? There are certain screen widths where the text pushes right up against the button

components/base/FooterCTA.tsx
```diff
@@ -0,0 +1,54 @@
...
+        <H2 isDark>{topic}</H2>
```
Finholt, line 59: Looks like the margins on this standard h2 are bumping the text off of the vertical center of the component and adding a lot of space between the header and sub-header compared to the designs

components/base/FooterCTA.tsx
```diff
@@ -0,0 +1,54 @@
...
+  background-color: ${PURPLE_DARK};
+  display: flex;
+  justify-content: space-between;
+  align-items: center;
+  @media (${maxWidth(BREAKPOINT_MOBILE)}) {
+    padding: 10px;
+    flex-direction: column;
+  }
+  /* Text sizes decrease in smaller screens */
+  > div > h2 {
```
Finholt: This definitely wouldn't be something to address in this ticket, but I wonder if we could update the typography components to take in a `customStyles` prop or something 🤔

components/base/FooterCTA.tsx
ShreyaThapa: Ah, good catch! 
For this, I ended up setting a `padding-right: 20px` on all screens except mobile.

components/base/FooterCTA.tsx
ShreyaThapa: I ended up unsetting the default margins, and added a bottom margin to the elements. This allows there to be a margin between the bottom text and the button on mobile too.

components/base/FooterCTA.tsx
ShreyaThapa: Yeah, that sounds good! I believe we want to revisit Typography later, so will add this as an action item there as well! Thanks, Connor!

Finholt (approved):

PR #44, opened 2020-04-29
WEB-670: Moving images folder and import SVGs as inline components

Finholt:


components/base/select/Select.tsx
```diff
@@ -1,8 +1,8 @@
 import React from "react";
 import ReactSelect, { components } from "react-select";
-import arrowSelectIcon from "../../../public/images/arrow-select-icon.svg";
-import DoubleArrowSelectIcon from "../../../public/images/double-arrows-select-icon";
-import CheckMarkIcon from "../../../public/images/check-mark-icon";
+import dropdownArrowSingle from "../../../assets/images/component-icons/dropdown-arrow-single.svg";
+import { ReactComponent as DropdownArrowDouble } from "../../../assets/images/component-icons/dropdown-arrow-double.svg";
```
Finholt, line 4: This is a good example of the two ways to import SVGs now. The top is how you would import it if you wanted to use it as the src of an img tag, and the bottom is how you would import it if you wanted to use it as an inline SVG

components/base/select/Select.tsx
```diff
@@ -27,7 +27,7 @@ const DefaultDropdownIndicator = (props: any) => {
 const CodeDropdownIndicator = (props: any) => {
   return (
     <components.DropdownIndicator {...props}>
-      <DoubleArrowSelectIcon />
+      <DropdownArrowDouble width={7} />
```
Finholt, line 30: Note that the width (and height) props are applied to the `<svg>` tag when everything is rendered

custom.d.ts
```diff
@@ -0,0 +1,6 @@
+declare module "*.svg" {
+  import React = require("react"); // eslint-disable-line no-unused-vars
+  export const ReactComponent: React.SFC<React.SVGProps<SVGSVGElement>>;
+  const src: string;
+  export default src;
+}
```
Finholt, line 6: This is so that TypeScript accepts SVG files can be imported as either a src string or a react component

Finholt (commented):
This looks bigger than it is; most of the changes are from moving the images folder to `/assets` and running our SVGs through [SVGOMG](https://jakearchibald.github.io/svgomg/), and then tweaking imports, etc. following that.

I also added two sub-folders to the images to categorize our icons between `content-icons` (icons that will be imported into the MDX and passed as props to the components, like those orange icons on the cards component) and `component-icons` (icons used directly within the component like arrows)

The really cool change here is using SVGR to allow us to import SVGs inline as React components (or as img src's like we have been)

components/base/Card.tsx
```diff
@@ -96,6 +95,12 @@ const LinkStyle = styled.div`
   position: absolute;
   top: 8px;
   right: 8px;
+
+  svg {
```
ShreyaThapa, line 99: Nice! 👍

ShreyaThapa (approved):
Nice touch with the comments showing examples of the new ways to import images. 😉 👍

PR #45, opened 2020-04-29
WEB-603: Add Netlify _redirects file as a proof of concept

sddaniels:
I'm creating a PR so I can get a deploy preview. I might play with it a little to see what I can get away with in terms of rules. I'll post another comment here when I have my final check in ready to review.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -25,6 +25,7 @@ packakge-lock.json
 # misc
 .DS_Store
 .env*
+.idea
 
 # debug
 npm-debug.log*
--- a/public/_redirects
+++ b/public/_redirects
@@ -1,0 +1,2 @@
+# redirect from                          to
+/guides/receive-money/onboarding.html    /docs/guides/receive-money/onboarding


sddaniels:
@sausman @Finholt @ShreyaThapa This is ready to review. The example redirect I checked in is functional in the deploy preview.

I tried playing around with pattern matching ("splats" according to Netlify) but they only work as the very last part of the URL you want to match. So making a pattern for all `*.html` paths won't work. Each page will need its own rule unfortunately.

Finholt (approved):
Love how simple Netlify makes this!

One thing to note, the way this is currently, it does redirect from `/guides/receive-money/onboarding.html`, but it does not redirect from `/guides/receive-money/onboarding` (sans `.html`)

I'm fine with that given it's a POC though

sddaniels:
> I'm fine with that given it's a POC though

Yeah I was only planning on adding one rule this go around to prevent re-work. For the paths that don't have the `.html` ending we might be able to use the splats for a single rule though. Will have to play with that again when we get to it.

PR #48, opened 2020-05-05
Web 643/Image download + Tooltip

ShreyaThapa:
- Used react-tippy for the Tooltip in images. https://bundlephobia.com/result?p=react-tippy@1.3.4
- Added a Tooltip component
- Added a ImageDownload component that wraps images that need a download link


components/base/Tooltip.tsx
```diff
@@ -0,0 +1,65 @@
...
+          }
+          .tippy-popper[x-placement^="right"] [x-arrow] {
+            border-right-color: ${GREY_8};
+          }
+          .tippy-popper[x-placement^="left"] [x-arrow] {
+            border-left-color: ${GREY_8};
+          }
+        `}
+      ></Global>
+      <ReactTooltip
```
ShreyaThapa, line 53: I've only included a few props from what's available in the library. We'll likely only need these few for our purposes, but we can add support for more if need be.

sausman (approved):

components/base/ImageDownload.tsx
```diff
@@ -0,0 +1,55 @@
...
+        <DownloadIconWrapper>
+          <Tooltip text={tooltipText} hasArrow position="bottom" distance={11}>
+            <a href={children} download={downloadFileName || ""}>
+              <DownloadIcon />
+            </a>
+          </Tooltip>
+        </DownloadIconWrapper>
+      </DownloadDiv>
+      <ImageDiv>
+        <img src={children} alt="" />
```
Finholt, line 52: Can we give this a `max-width: 100%` or something so it shrinks on smaller mobile screens?

ShreyaThapa: Good point! Totally overlooked that 😬

ShreyaThapa: This [commit](https://github.com/Dwolla/dev-portal/pull/48/commits/6a8d1f2db59c9945e35f9aed498cadcb9599b596) should take care of that!

Finholt (approved):

PR #49, opened 2020-05-05
WEB-672: Implementing Guide Pagination

Finholt:


modules/pages.ts
```diff
@@ -42,7 +42,7 @@ const Pages = {
   },
 
   under(path: string): Page[] {
-    const normalizedPath = `/${path.replace(/^\//, "").replace(/\/$/, "")}/`;
+    const normalizedPath = `/${path.replace(/^\//, "").replace(/\/$/, "")}`;
```
Finholt, line 45: Removed trailing `/` in order to return the index file under the path as well

next.config.js
```diff
@@ -6,8 +6,8 @@ const { PHASE_DEVELOPMENT_SERVER } = require("next/constants");
 module.exports = (phase) =>
   withImages(
     withMdxEnhanced({
-      // layoutPath: 'layouts',
-      // defaultLayout: true,
+      layoutPath: "layouts",
+      defaultLayout: true,
```
Finholt, line 10: Applies `layouts/index.tsx` by default if no layout is specified in the frontmatter

layouts/guides.tsx
```diff
@@ -0,0 +1,51 @@
...
+              nextHref={nextPage ? nextPage.id : ""}
+            />
+          </StyledPagination>
+        )}
+
+        <FooterCTA
+          topic="Test in the Sandbox for free today."
+          description="Use sandbox environment to test API requests."
+          button={{ text: "Get API Keys" }}
+        />
```
Finholt, line 47: Using these layouts, we gain access to the frontMatter and can add content across several pages. In this case, I'm adding the pagination and footer cta to all pages with `layout: guides` in their frontmatter

sausman (approved):

PR #50, opened 2020-05-11
Add CodeExamples component

sausman:


components/partial/code/CodeExamples.tsx
```diff
@@ -0,0 +1,154 @@
...
+  border-bottom-left-radius: 5px;
+  border-bottom-right-radius: 5px;
+
+  > pre {
+    margin: 0;
+
+    > code {
+      color: ${GREY_3};
+
+      .hljs- {
```
sausman, line 46: Wasn't quite sure which colors to use for what, so these may be something to revisit.

components/partial/code/CodeExamples.tsx
```diff
@@ -0,0 +1,162 @@
...
+import highlight from "../../../modules/highlight";
+import Select from "../../base/select/Select";
+
+const SwitcherBar = styled.div`
+  background-color: ${PURPLE_DARKER};
+  padding: 10px 8px;
+  border-top-left-radius: 5px;
+  border-top-right-radius: 5px;
+  display: flex;
+`;
```
Finholt, line 31: This is missing the copy button and the designs has it right-aligned

components/partial/code/CodeExamples.tsx
```diff
@@ -0,0 +1,162 @@
...
+  // selectedLanguage is one of the context's languageOptions
+  useEffect(() => {
+    const selectedLanguageOptionsFromCtx = getLanguageOptions(
+      selectedLanguage,
+      ctx.languageOptions
+    );
+    if (selectedLanguageOptionsFromCtx) {
+      ctx.setSelectedLanguage(selectedLanguageOptionsFromCtx);
+    }
+  }, [selectedLanguage]);
```
Finholt, line 123: I noticed that if there is a code block on a page that has an option that is one of the context's languageOptions, but not the selected context language option, it does change the context when the page loads, which forces the user to switch it back to their preferred option

Example: [Go to this page](https://deploy-preview-50--dwolla-dev-portal.netlify.app/docs/guides/webhooks/create-a-webhook-subscription/create-subscription) (JS should be the selected language by default) and then use the pagination to go to the next page. The selected language changes to Ruby because that is the only available language for one of the code blocks

I'm not entirely sure how everything _should_ behave in this scenario, but this feels like it could be frustrating as it is

Finholt (approved):
I did notice a couple of things that could be improved, but I don't want to hold this PR up since I think it could help Spencer with his content updates. I'm fine merging this in as is and addressing them with a future commit!

PR #53, opened 2020-05-12
Tab bar panel component

ShreyaThapa:
After trying out a couple of existing react-components for Tab Panel functionality, I ended up coding it from scratch. It was much easier for me to style 😅


components/base/TabBarPanel.tsx
```diff
@@ -0,0 +1,115 @@
...
+  align-items: center;
+  justify-content: center;
+  &.active {
+    background-color: ${WHITE_PRIMARY};
+    border-color: ${GREY_2};
+    border-style: solid;
+    border-width: 0 1px 0 1px;
+    border-radius: 5px 5px 0 0;
+    padding-top: 16px;
+    :nth-of-type(1) {
```
ShreyaThapa: Styling the border of the first and las elements of the Tab Bar.

components/base/__tests__/TabBarPanel.test.tsx
```diff
@@ -0,0 +1,54 @@
...
+  {
+    label: "node",
+    icon: nodeGray,
+    iconActive: nodeGray,
+    content: "Here's some content related to Node.",
+  },
+  {
+    label: "ruby",
+    icon: rubyGray,
+    iconActive: rubyColor,
```
ShreyaThapa, line 21: This is an example of passing both the gray and the colored icons to the component. Since we don't have the colored icons for the other languages yet, I ended up passing the same gray icons twice for now.

components/base/TabBarPanel.tsx
```diff
@@ -0,0 +1,115 @@
...
+import { ROBOTO } from "../typography";
+
+// Styles
+const Container = styled.div`
+  max-width: auto;
+  border-radius: 5px;
+  background-color: ${WHITE_PRIMARY};
+  :hover {
+    box-shadow: ${BOX_SHADOW_7};
+  }
```
Finholt: Having a hover state on this entire container looks off to me since it'll take up a pretty large portion of the page in practice

components/base/TabBarPanel.tsx
```diff
@@ -0,0 +1,115 @@
...
+  box-sizing: border-box;
+  border-radius: 5px 5px 0 0;
+  border-color: ${GREY_2};
+  border-style: solid;
+  border-width: 1px 1px 0 1px;
+  background-color: ${GREY_1};
+  display: flex;
+  flex-wrap: nowrap;
+  /* Tab Bar will scroll on smaller screens */
+  overflow-x: auto;
```
Finholt: Nice!

components/base/TabBarPanel.tsx
```diff
@@ -0,0 +1,115 @@
...
+  &::-webkit-scrollbar {
+    display: none;
+  }
+`;
+
+const TabStyle = styled.div`
```
Finholt: Can we make the cursor a pointer when hovering over an inactive tab?

components/base/TabBarPanel.tsx
```diff
@@ -0,0 +1,115 @@
...
+            className={classnames({ active: tab === activeTab })}
+            key={tab.label}
+            onClick={() => {
+              setActiveTab(tab);
+            }}
+          >
+            <StyledIcon
+              src={tab === activeTab ? tab.iconActive : tab.icon}
+              alt={tab.label}
+            />
```
Finholt: I think we'll eventually need to accommodate a version of this that uses something besides icons, like we see on the [changelog page](https://projects.invisionapp.com/d/main#/console/19172953/400098965/preview), but I can't remember if the tabs on that page were something we deprioritized or not, so this might be good for now

components/base/TabBarPanel.tsx
```diff
@@ -0,0 +1,115 @@
...
+    <Container>
+      <StyledTabBar>
+        {tabs.map((tab) => (
+          <TabStyle
+            className={classnames({ active: tab === activeTab })}
+            key={tab.label}
+            onClick={() => {
+              setActiveTab(tab);
+            }}
+          >
```
Finholt: This is really slick with a mouse, but I have some concerns about accessibility since it can't be tabbed to accessed with a keyboard

Finholt (commented):
I noticed a couple of things to address, but overall, this is looking really great!

components/base/TabBarPanel.tsx
ShreyaThapa: I agree. I went off of the mocks, but I think we can skip the hover state for the container. 👍

components/base/TabBarPanel.tsx
ShreyaThapa: Good catch! Ughh these things that I keep missing!

components/base/TabBarPanel.tsx
ShreyaThapa: Yeah I though about that too. For now, I tried to keep it simple which works for the SDKs panel, but keeping in mind that there may be other variations that we may come across, in which case we can revisit the component!

components/base/TabBarPanel.tsx
ShreyaThapa: That's a good thought. I'll look into making it keyboard accessible!

components/base/TabBarPanel.tsx
```diff
@@ -1,62 +1,12 @@
+/* eslint-disable react/no-array-index-key */
 import React, { useState } from "react";
+import { Tab, Tabs, TabList, TabPanel } from "react-tabs";
```
ShreyaThapa, line 3: Using react-tabs package which comes with accessibility built in.

components/base/TabBarPanel.tsx
ShreyaThapa: Heya! This [commit](https://github.com/Dwolla/dev-portal/pull/53/commits/1e51e9f7118d82bd9cfb9c53e85455e72737d47e) should address all comments.

Finholt (approved):
Looks great, Shreya!

One slight concern I have is when clicking on a tab for the first time (i.e. incognito window), there was a small (but noticeable) delay of up to half a second before the colored logo loaded in. I'm hoping that this is caused by deploy previews having less resources or something with storybook, but we may want to come back later and optimize the svgs and/or pre-load them when the component loads if this happens in prod. Nothing worth holding up this ticket for though!

ShreyaThapa:
That's a good point, Connor! I didn't think of checking that. But definitely a good thing to look out for in Prod! 👍 Thanks!

PR #54, opened 2020-05-12
Added oarnge variant to Badge

ShreyaThapa:
Made variant optional and added "default" and "orange" variants, but let me know if something else sounds better than "orange"!

Examples,
- with variant:  `<Badge text="some text or number" variant="orange" />`
- without variant:  `<Badge text="some text" />`

### DIFF
--- a/components/base/Badge.tsx
+++ b/components/base/Badge.tsx
@@ -1,13 +1,8 @@
 import React from "react";
 import styled from "@emotion/styled";
-import { GREY_5, GREY_4 } from "../colors";
+import classnames from "../../modules/classnames";
+import { GREY_5, GREY_4, WHITE_PRIMARY, ORANGE_PRIMARY } from "../colors";
 import { POPPINS } from "../typography";
-
-type Props = { text: string };
-
-function Badge({ text }: Props) {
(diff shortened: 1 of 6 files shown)


components/base/Badge.tsx
```diff
@@ -24,6 +19,25 @@ const StyledBadge = styled.div`
...
+    height: 22px;
+    padding: 2px 8px;
+  }
 `;
 
+type Props = { text: string; variant?: "default" | "orange" };
+
+function Badge({ text, variant }: Props) {
+  return (
+    <StyledBadge className={classnames({ orange: variant })}>
```
Finholt: I think we just want this just be `className={ variant }`

As it is now, it'll apply the `.orange` class whenever variant is truthy, which is the case even if variant is set to "default"

Finholt (commented):
I think "orange" is perfect for the variant name. It describes what it is and leaves the door open for other variant styles if they need to be added in the future

components/base/Badge.tsx
ShreyaThapa: Ah, good catch! 👍

Finholt (approved):

PR #55, opened 2020-05-13
WEB-659: Populating Guide index page using frontmatter

Finholt:


modules/images.import.ts
```diff
@@ -0,0 +1,13 @@
+const importAllImagesFromContext = (r) => {
+  const images = {};
+  r.keys().forEach((filename) => {
+    images[filename.replace("./", "")] = r(filename);
+  });
+  return images;
+};
+
+export const contentIcons = importAllImagesFromContext(
+  require.context("../assets/images/content-icons", false, /\.(png|jpe?g|svg)$/)
```
Finholt, line 10: This creates an object that allows us to access all of the images within this directory. I'm doing this because imports seem to be completely unavailable in the frontmatter.

The more I look at this, the more I'm wondering if this is needlessly complicated, and maybe we should move content-images back to the public folder (which kills me to say since I moved them out of there to begin with 🤦). Component-images _could_ stay in assets, but at that point, are there even any substantial benefits for doing so? Thoughts?

pages/docs/guides/index.mdx
```diff
@@ -6,69 +6,28 @@ description: "Programmatically send bank transfers with our developer API."
...
-    icon: guidesIcon,
-    category: "Prerequisites",
-    href: "/docs/guides/migrate-to-v2",
-  },
-];
+      "Step-by-step instructions to get you set up in the Dwolla API",
+  };
+  const guideOverviewPages = Pages.under("docs/guides").filter(
+    (guide) => !!guide.guide && guide.guide.step === 0
+  );
```
Finholt, line 20: Getting all guides whose `step===0`, which is the case for the "overview" pages

Finholt (commented):
Before I close out the ticket overall, I want to wait for the content updates that Spencer is working on, but this seemed like a pretty good piece that I could get out there sooner

modules/images.import.ts
sausman: Yeah that's a tough one, curious what everyone else thinks. I like the changes you made to how we're doing images, but I'm not strongly against moving them back.

sausman (approved):

modules/images.import.ts
Finholt: Perhaps I'm overthinking it then. Since this gets the job done, I'll go ahead and merge it in as is. We can always reevaluate later if it becomes too confusing or frustrating while working with the content

PR #56, opened 2020-05-13
Improve CodeExamples

sausman:
- Support single, plaintext example
- Add copy button
- Add `useCopy` hook


components/util/useCopy.tsx
```diff
@@ -0,0 +1,26 @@
+import { useEffect, useState } from "react";
```
sausman: Where do you guys think we should put custom hooks? Not sure this is the best/most obvious place.

Finholt: I agree that the util folder probably isn't quite the right place. In the demo app, we had them on them in a hooks folder on the same level as the components folder, but I could also see putting them in a hooks folder inside the components folder

components/partial/code/CodeExamples.tsx
```diff
@@ -144,19 +105,31 @@ export default function CodeExamples({
...
+        {onlyOneExample ? (
+          <BarText>{activeLanguage}</BarText>
+        ) : (
+          <Select
+            autoWidth
+            variant="code"
+            options={options}
+            selectedValue={options.find((o) => o.value === activeLanguage)}
+            setSelectedValue={({ value }) => setSelectedLanguage(value)}
+          />
```
Finholt, line 124: I noticed that one of the options is `bash`, but we have it as `raw` in the dropdown that we have in the top nav. Can we switch one to the other so it's more consistent?

sausman: Yeah nice catch,

I updated the value in `LANGUAGE_OPTIONS` for now to be `bash`, but we may want to update the .mdx files to use ```raw. cc @spencerhunter

components/util/useCopy.tsx
sausman: Ended up moving to `/hooks`

Finholt (approved):

PR #57, opened 2020-05-14
WEB-610: Add breakpoints and refactor to make them clearer to use

sddaniels:
This PR implements the breakpoints as Bootstrap-style functions that tell you a bit more about which screens you are targeting in both the up and down directions. It should be a bit safer to use and require fewer imports.

### DIFF
--- a/components/base/Banner.tsx
+++ b/components/base/Banner.tsx
@@ -3,7 +3,7 @@ import classnames from "../../modules/classnames";
 import Button from "./Button";
 import { PURPLE_DARK, WHITE_PRIMARY, GREY_4 } from "../colors";
 import { POPPINS, ROBOTO } from "../typography";
-import { maxWidth, BREAKPOINT_IPAD, BREAKPOINT_MOBILE } from "../breakpoints";
+import { breakDown } from "../breakpoints";
 
 // Styled Banner component
 const StyledBanner = styled.div`
@@ -34,11 +34,11 @@ const StyledTopic = styled.div`
   font-weight: 200;
   line-height: 52px;
   text-align: center;
-  @media (${maxWidth(BREAKPOINT_IPAD)}) {
+  @media (${breakDown("sm")}) {
     font-size: 30px;
     line-height: 46px;
   }
-  @media (${maxWidth(BREAKPOINT_MOBILE)}) {
+  @media (${breakDown("xs")}) {
     font-size: 22px;
     line-height: 33px;
   }
@@ -54,7 +54,7 @@ const StyledDescription = styled.div`
   &.filters {
     margin-bottom: 52px;
   }
-  @media (${maxWidth(BREAKPOINT_IPAD)}) {
+  @media (${breakDown("sm")}) {
     font-size: 16px;
     line-height: 20px;
   }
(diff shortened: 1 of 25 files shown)


components/base/Banner.tsx
```diff
@@ -34,11 +34,11 @@ const StyledTopic = styled.div`
   font-weight: 200;
   line-height: 52px;
   text-align: center;
-  @media (${maxWidth(BREAKPOINT_IPAD)}) {
+  @media (${breakDown("sm")}) {
```
sddaniels, line 37: This applies a breakpoint downwards using `max-width`, with the rule starting to apply with the "sm" screen size.

components/base/FilterTabs.tsx
```diff
@@ -43,7 +43,7 @@ export const TabStyle = styled.div`
 const SelectWrapper = styled.div`
   width: 150px;
   padding-bottom: 10px;
-  @media (${minWidth(BREAKPOINT_MOBILE)}) {
+  @media (${breakUp("sm")}) {
```
sddaniels, line 46: This applies a breakpoint upwards using `min-width`, with the rule starting to apply on the "sm" screen size.

components/breakpoints.ts
```diff
@@ -1,6 +1,49 @@
...
+// > <: extra large devices - large desktops, etc.
+const BREAKPOINT_5 = 1600;
+// >: extra extra large devices - Big McLargeHuge devices
 
-export const BREAKPOINT_DESKTOP = 1024;
-export const BREAKPOINT_IPAD = 768;
-export const BREAKPOINT_MOBILE = 576;
+function breakUpPx(breakpoint) {
+  switch (breakpoint) {
+    case "sm":
```
sddaniels, line 17: No `xs` size defined or needed here, because if you are using `min-width` then XS is included by using standard CSS.

components/breakpoints.ts
```diff
@@ -1,6 +1,49 @@
...
+      return BREAKPOINT_1;
+    case "sm":
+      return BREAKPOINT_2;
+    case "md":
+      return BREAKPOINT_3;
+    case "lg":
+      return BREAKPOINT_4;
+    case "xl":
+      return BREAKPOINT_5;
+    default:
```
sddaniels, line 44: No `xxl` size defined or needed here, because if you are using max-width then XXL is included by using standard CSS.

components/breakpoints.ts
```diff
@@ -1,6 +1,49 @@
...
+      return breakpoint;
+  }
+}
+
+export const breakUp = (breakpoint) => `min-width: ${breakUpPx(breakpoint)}px`;
```
sausman: Nice! I like how we don't have to import the breakpoint anymore and can just pass a string.

We might also want to add a type to `breakpoint` to ensure we pass a valid breakpoint to it. Something like:

```typescript
type Breakpoint = "xs" | "sm" | "md" | "lg" | "xl" | "xxl"

export const breakUp = (breakpoint: Breakpoint) => `min-width: ${breakUpPx(breakpoint)}px`;
```

Finholt (approved):

components/breakpoints.ts
sddaniels: I just pushed up a commit that adds types. It does remove the ability to pass in a custom width, which I was sort of intending to support if possible, but I think it's ok without.

sausman (approved):

PR #59, opened 2020-05-14
WEB-659: Refactoring main content area layout

Finholt:


layouts/guides.tsx
```diff
@@ -6,10 +6,14 @@ import FooterCTA from "../components/base/FooterCTA";
...
-  margin: 20px;
+  padding: 20px;
 
   @media (${breakUp("md")}) {
-    margin: 40px;
+    padding: 40px;
+  }
+
+  @media (${breakUp("lg")}) {
+    width: 70%;
```
Finholt, line 16: The main point of this PR is to allow these layouts, and the components in them, to not be bound by width the mdx content has to accommodate the "on this page" component. I.e. Components like the FooterCTA here can take up the full width of the content area

layouts/index.tsx
```diff
@@ -25,15 +36,19 @@ export const DefaultMDXWrapper = ({ children, frontMatter }: Props) => (
...
+        <AnchorsSetter>
+          {groupCodeExamples({
+            children,
+            into: CodeExamples,
+          })}
+        </AnchorsSetter>
+      </MDXStyleWrapper>
+    </ContentWrapper>
+
+    <OnThisPage />
```
Finholt, line 50: This previously appeared in an `OnThisPageWrapper`, but that wrapper's height was causing this to scroll off the screen prematurely, so I also did some additional refactoring with that

sausman (approved):

PR #60, opened 2020-05-15
WEB-680: Updating Footer Layout

Finholt:


components/layout/Footer.tsx
```diff
@@ -2,45 +2,87 @@ import styled from "@emotion/styled";
...
+
+const Copyright = styled.div`
+  font-family: ${ROBOTO};
+  font-size: 12px;
+  line-height: 21px;
+  color: ${PARAGRAPH_TEXT};
+  margin-top: 14px;
+`;
+
+const LinkContainer = styled.div`
```
ShreyaThapa, line 45: Could we do something here to allow the footer links to stick more to the left side of the screen (closer to the logo) than the right side? This looks great on a regular desktop and smaller screens, but when the browser stretched out, the links stick to the right leaving some white space between the logo and the link.

Finholt: The main thing I was trying to avoid was setting an arbitrary max-width, but you're right that it does look off. My latest commit should address it by making the max-width equal to the width of the content area when the OnThisPage component is present. Thanks for having me take another look!

ShreyaThapa: Thanks, Connor. That looks neat! 👍

ShreyaThapa (approved):

PR #63, opened 2020-05-18
Add table styles

sausman:


components/base/Typography.tsx
```diff
@@ -136,6 +137,50 @@ export const listStyles = css`
...
+      line-height: 17px;
+      padding: 8px 16px;
+    }
+  }
+
+  tbody {
+    tr {
+      box-shadow: 0 0 0 1px ${GREY_2}; /* https://stackoverflow.com/a/48831092 */
+
+      &:nth-of-type(1) {
```
sausman: I initialled had `:first-child` here, but Emotion was complaining about it potentially being an issue with SSR (🤷) so I changed it to this.

components/base/Typography.tsx
```diff
@@ -136,6 +137,50 @@ export const listStyles = css`
   }
 `;
 
+// Table styles
+export const tableStyles = css`
```
Finholt: Could we export these from `Table.tsx` instead since they're not really typography? I did something similar with `imageStyles` not too long ago

components/base/Typography.tsx
```diff
@@ -136,6 +137,50 @@ export const listStyles = css`
...
+        background: ${GREY_1};
+      }
+    }
+
+    td {
+      font-family: ${ROBOTO};
+      font-size: 14px;
+      line-height: 20px;
+      color: ${PARAGRAPH_TEXT};
+      padding: 10px 16px;
```
Finholt: Can we bump up the vertical padding a bit more? It feels pretty tight compared to the designs

components/base/Typography.tsx
Finholt: I believe the designs have tables taking up the full width of the content area whenever they're used. Is that something we could build in here? Or do you think there are some situations where it would be better only be the width of the content?

Finholt: I noticed that the tables do not overflow gracefully on smaller screens ([example](https://deploy-preview-63--dwolla-dev-portal.netlify.app/docs/guides/migrate-to-v2/05-v1-to-v2-endpoint-changes#v1-to-access-api-endpoint-changes))

Finholt: Related, the current styling implemented in that file for the component version of tables is different than what's here. Can we make them consistent with each other?

stories/base/Typography.stories.tsx
```diff
@@ -50,6 +51,51 @@ const orderedlList = [
   { id: "2", text: "Ordered list text goes here" },
   { id: "3", text: "Ordered list text goes here" },
 ];
+const tableContents = (
+  <>
+    <thead>
+      <tr>
+        <th align="left">Bank to Dwolla Network Clearing</th>
```
Finholt: I don't believe alignment is something that you can specify in markdown (correct me if I'm wrong!), so should we left align the header cells by default?

spencerhunter: You should be able to specify alignment in tables with markdown. Example below

```
| Default Header | Left Align | Right Align | Center Align |
| --- | :-- | --: | :-: |
```

Finholt: The More You Know! Thanks Spencer! Disregard this comment then

components/base/Table.tsx
```diff
@@ -40,21 +40,31 @@ const StyledTable = styled.div`
...
-  }
-  /* Bottom-right td cell */
-  tr:nth-last-of-type(1) td:nth-last-of-type(1) {
-    border-radius: 0 0 5px 0;
-  }
-  /* Bottom-left td cell */
-  tr:nth-last-of-type(1) td:nth-of-type(1) {
-    border-radius: 0 0 0 5px;
+
+  /* td border-radius */
```
sausman, line 44: This is the only thing I really changed here. The border radius for the top row of the `<tbody>` wasn't working on some of the examples in the docs.

Finholt (approved):

PR #64, opened 2020-05-20
WEB-674: Updating Side Nav Hover and Active Styles

Finholt:


pages/_app.tsx
```diff
@@ -7,36 +7,36 @@ import { AnchorsProvider } from "../components/util/Anchors";
...
 import AuthPage from "../components/AuthPage";
 import fetcher from "../modules/fetcher";
-import homeIcon from "../assets/images/component-icons/side-nav/home-nav-icon.svg";
-import guidesIcon from "../assets/images/component-icons/side-nav/guides-nav-icon.svg";
-import conceptsIcon from "../assets/images/component-icons/side-nav/concepts-nav-icon.svg";
-import sdksToolsIcon from "../assets/images/component-icons/side-nav/sdks-tools-nav-icon.svg";
+import { ReactComponent as HomeIcon } from "../assets/images/component-icons/side-nav/home-nav-icon.svg";
+import { ReactComponent as GuidesIcon } from "../assets/images/component-icons/side-nav/guides-nav-icon.svg";
+import { ReactComponent as ConceptsIcon } from "../assets/images/component-icons/side-nav/concepts-nav-icon.svg";
+import { ReactComponent as SdksToolsIcon } from "../assets/images/component-icons/side-nav/sdks-tools-nav-icon.svg";
```
Finholt, line 13: Using inline SVGs so we can change the fill color when `.active` class is applied

components/layout/SideNav.tsx
```diff
@@ -171,12 +201,15 @@ const groupToggleDocLinkStyles = css`
...
+  font-family: ${ROBOTO};
   color: #52627b;
-  padding: 4px 0 4px 10px;
+  padding: 4px 20px 4px 10px;
+  border-left: 1px solid transparent;
+  margin-left: -1px;
 
   &:hover {
-    color: ${ORANGE_PRIMARY};
+    border-left: 1px solid ${GREY_6};
```
Finholt, line 211: This wasn't in Kenny's mock-ups, but since the normal border color is the same as the hover background color, it felt like it needed something. Open to discussion!

components/layout/SideNav.tsx
```diff
@@ -186,17 +219,48 @@ const GroupToggle = styled.div`
...
+};
+
+const DocGroup = ({ title, docs }: DocGroupProps) => {
+  const { pathname } = useRouter();
+  const [isExpanded, setIsExpanded] = useState(
+    !!docs.find((d) => d.id === pathname)
+  );
+  useEffect(() => {
+    setIsExpanded(docs.find((d) => d.id === pathname) ? true : isExpanded);
+  }, [pathname]);
```
Finholt, line 240: Tracking whether a group is expanded at the group level rather than the side-nav level. This made it significantly easier to make sure the group containing the active doc is expanded when accessing from an index page or on refresh

components/layout/SideNav.tsx
```diff
@@ -239,114 +295,68 @@ function DocLink(props: DocLinkProps) {
...
-  }, [selectedSectionHref]);
-
-  const mostRecentSection = props.links.find(
-    hrefIs(state.mostRecentSectionHref)
+  const [activeSection, setActiveSection] = useState(
+    findSelectedSection(sectionLinks, pathname)
   );
+  useEffect(() => {
+    setActiveSection(findSelectedSection(sectionLinks, pathname));
+  }, [pathname]);
```
Finholt: With expanded state handled in the children, using a reducer seemed excessive, so I refactored to use the useState hook instead

Finholt (commented):
It might be helpful to view the diffs of my two commits separately with this one. The first one covers most of the style changes. The second covers functionality to automatically expand the group containing the active doc as well as some refactoring and cleanup

sausman (approved):
👍

PR #65, opened 2020-05-22
add card descriptions; move content icons to content images directory

spencerhunter:


pages/guides/dwolla-js/index.mdx
```diff
@@ -1,13 +1,6 @@
 ---
 layout: guides
 title: "Overview"
-group:
-  category: Funding Sources
-  title: Dwolla.js
-guide:
-  step: 0
-  filter: get-building
-  icon: guides-icon.svg
```
Finholt, line 10: Was this the file we were talking about leaving out for now until we have a good way to implement it? If so, we could remove it now since it's causing problems, but I'm also fine waiting for another of the content tickets to follow up since we haven't fully discussed a game plan yet

Finholt (approved):

pages/guides/dwolla-js/index.mdx
spencerhunter: Ah, yes good catch. I can remove that file in the cleanup ticket I'm picking up next. May discuss a few options with the team tomorrow on this topic.

PR #66, opened 2020-05-22
Added ExternalCTA component

ShreyaThapa:


components/base/ExternalCTA.tsx
```diff
@@ -0,0 +1,87 @@
...
+      fill: ${GREY_4};
+    }
+  }
+`;
+
+type Props = {
+  icon: string;
+  title?: string;
+  description?: string;
+};
```
Finholt: What are your thoughts on adding an "href" prop and implementing the link within this component? We certainly don't have to (the "card" component, for instance, doesn't have one and we just add the link as a parent if needed), but it might make using this component a bit simpler since it looks like it will always be used as an external link

assets/images/content-icons/developer-community-icon.svg
```diff
@@ -0,0 +1,18 @@
...
+                    <path d="M40,14 C41.1045695,14 42,13.1045695 42,12 C42,10.8954305 41.1045695,10 40,10 C38.8954305,10 38,10.8954305 38,12 C38,13.1045695 38.8954305,14 40,14 Z" id="Path"></path>
+                    <path d="M32,14 C33.1045695,14 34,13.1045695 34,12 C34,10.8954305 33.1045695,10 32,10 C30.8954305,10 30,10.8954305 30,12 C30,13.1045695 30.8954305,14 32,14 Z" id="Path"></path>
+                    <path d="M24,14 C25.1045695,14 26,13.1045695 26,12 C26,10.8954305 25.1045695,10 24,10 C22.8954305,10 22,10.8954305 22,12 C22,13.1045695 22.8954305,14 24,14 Z" id="Path"></path>
+                </g>
+            </g>
+        </g>
+    </g>
+</svg>
```
Finholt: This is probably already on your radar, but just a reminder that depending on the merge order between this and Spencer's PR that changes the images folder structure, there might be some conflicts or weirdness here

components/base/ExternalCTA.tsx
ShreyaThapa: Yeah, that's a good idea for this card that always contains a link. 
I'll add an `href` prop and wrap the CTA in a `<Link>`! 👍

components/base/ExternalCTA.tsx
```diff
@@ -0,0 +1,92 @@
...
+type Props = {
+  icon: string;
+  title?: string;
+  description?: string;
+  href: string;
+};
+
+function ExternalCTA({ icon, title, description, href }: Props) {
+  return (
+    <Link href={href} passHref>
```
Finholt, line 77: One last small nitpick, can we have this open in a new tab since it's meant to be for external links?

ShreyaThapa: D'oh! Missed that! Very good catch!

ShreyaThapa: [This](https://github.com/Dwolla/dev-portal/pull/66/commits/c1e7b5d84d64b30443817ba6b7cd1efee3ca5847) should fix it!

assets/images/content-icons/developer-community-icon.svg
ShreyaThapa: Thanks for that reminder still! Re-arranged the image location [here](https://github.com/Dwolla/dev-portal/pull/66/commits/c1e7b5d84d64b30443817ba6b7cd1efee3ca5847).

Finholt (approved):
🙌

PR #68, opened 2020-05-27
Web 657/logo links

ShreyaThapa:
Added LogoLinks component
Updated TabBarPanel component to add hover states

What do you think of this structure for the new logos added in this PR?:

Screen Shot 2020-05-26 at 7 05 59 PM

I'm not really tied to it; open to any changes!


components/base/LogoLinks.tsx
```diff
@@ -0,0 +1,88 @@
...
+          <StyledLink
+            // eslint-disable-next-line react/no-array-index-key
+            key={index}
+            href={tab.href}
+            rel="noopener noreferrer"
+            target="_blank"
+            onMouseEnter={() => setActiveTab(tab)}
+            onMouseLeave={() => setActiveTab(null)}
+            onFocus={() => setActiveTab(tab)}
+            onBlur={() => setActiveTab(null)}
```
Finholt: Could we track focused and hovered tabs separately? I.e. instead of the one `activeTab`, there would be a `focusTab` and a `hoverTab` (or something like that). The reason for this is right now, if you tab a given logo and then hover the mouse over a different logo, the `activeTab` (and colored logo state associated with it) switches, but the focus remains on a now grayed out logo

It's definitely an edge case, but it shouldn't be a hard fix, especially since you're already doing something similar in the TabBarPanel

Finholt (commented):
I left a comment on one thing that I noticed as an edge case, but all in all, the component looks and works great! Nice work, Shreya!

components/base/LogoLinks.tsx
ShreyaThapa: Great catch Connor!

ShreyaThapa: Would something like [this](https://github.com/Dwolla/dev-portal/pull/68/commits/0a552e952da06bd28789ed0888627053e07a6c3d) work?

Finholt: Hell yeah!! That's awesome

Finholt (approved):

PR #71, opened 2020-05-27
WEB-682: Implementing Concepts

Finholt:


components/partial/MDXStyleWrapper.tsx
```diff
@@ -60,7 +60,7 @@ const MDXStyleWrapper = ({ children }: { children: any }) => (
         }
       }
 
-      a {
+      a:not([class]) {
```
Finholt, line 63: I was running into a lot of issues with these styles overriding links that were part of components

components/partial/IndexPageGrid.tsx
```diff
@@ -61,28 +69,32 @@ const IndexPageGrid = ({ unfilteredCards, bannerProps }: Props) => {
...
+              <FilterTabs
+                tabs={filters}
+                filter={
+                  router.query.filter ? router.query.filter.toString() : null
+                }
+                setFilter={setFilter}
+              />
+            ) : (
+              <></>
+            )
```
Finholt, line 82: Not rendering the FilterTabs if "All" is the only one

ShreyaThapa:
Looking good! And thanks for the extra fixes 🙌

ShreyaThapa (approved):

PR #72, opened 2020-05-28
Order categories and ungrouped doc links in SideNav

sausman:


pages/guides/_section.yml
```diff
@@ -0,0 +1,5 @@
+categories:
```
sausman, line 1: As I was working on this, it didn't feel quite right specifying Category ordering/weights within individual Docs' frontmatter. The Category order doesn't belong to an individual Doc, it's specific to a Section.

If data that belongs to a Section can be defined within any Doc, it's not obvious where that data lives/should live, and how to deal with potentially conflicting data in the many different places it could exist in.

We ran into a similar issue when it came to defining Groups of docs, and solved it by allowing a Group to be created  by adding `group:` to any `index.mdx` Doc's metadata.

This is similar, but allows a section to be created by adding a `_section.yml`. In the future we could expand what's defined here to include more of the stuff we currently have hardcoded in the app (if we want).

README.md
```diff
@@ -53,6 +53,10 @@ Page content
 <MyComponent />
 ```
 
+#### Sections
+
+`_section.yml`
```
sausman: (todo)

components/layout/SideNav.tsx
```diff
@@ -460,22 +467,22 @@ const SideNav = ({ sectionLinks, pages, mobileItems }: SideNavProps) => {
...
           <CategoriesWrap>
-            {keys(categories).map((c) => {
+            {sortCategories(activeSection.href, keys(categories)).map((c) => {
               const categoryDocs = categories[c];
               const groups = groupsFrom(categoryDocs);
 
               return (
                 <Category key={c}>
-                  {c && <CategoryHeading>{c}</CategoryHeading>}
+                  {c !== "undefined" && <CategoryHeading>{c}</CategoryHeading>}
```
sausman, line 476: This is to prevent `"undefined"` from being displayed as a category heading. Not pretty, but didn't want to spend much time on it since I think it'd make sense to do a SideNav cleanup ticket at some point.

ShreyaThapa: Sweet! The rogue UNDEFINED category is no more! 

However, not sure if it's related, but the `guides/dwolla.js/index.mdx` file that doesn't have a "group" or "weight" defined is floating to the top of the SideNav. Navigating to that file gives a `Cannot read property 'step' of undefined` error. Any way we could keep it from appearing in the SideNav? 

Screen Shot 2020-05-28 at 11 35 27 PM

pages/guides/_section.yml
ShreyaThapa: Agreed. I like this!

components/layout/SideNav.tsx
sausman: Yeah, it's doing the same on master (the error). Do you know if there's a category we want the Dwolla.js guide to display under?

ShreyaThapa: Yeah, it'd be under the category: FUNDING SOURCES. I believe it already is, but this particular doc doesn't have a group or step defined.

Finholt: I think that's supposed to be a shared index page that's going away until we have some way to handle that situation: https://dwolla.slack.com/archives/GQH89J8G0/p1590505702004800

ShreyaThapa (approved):

components/layout/SideNav.tsx
```diff
@@ -460,22 +467,22 @@ const SideNav = ({ sectionLinks, pages, mobileItems }: SideNavProps) => {
...
 
-                  {groups.map((g) => {
-                    const docs = categoryDocs.filter(byDocGroup(g));
-                    const ungroupedGroup = g === null;
+                  {sortByWeight(groups).map((g?: PageGroup) => {
```
Finholt, line 478: Maybe I'm just a little bit blind or not focusing enough, but can you explain how I should add the `weight` to groups to sort them?

sausman: Yeah sorry for not including that! You'd add `weight` underneath `group` in a doc that defines a group (`index.mdx` file). For example:

image

It might make sense in the future to split out the group definition into `_group.yml` (similar to how we're doing `_section.yml`) to make things clearer/more consistent.

Finholt (approved):
Thanks for answering that question, and sorry I got distracted in the middle of looking at this, but it looks great to me!

PR #73, opened 2020-05-28
Web 658/homepage

ShreyaThapa:
Changed homepage file type to .mdx and added front-matter.
Updated Banner to use `"width: 100%"` and added a customer width to <BannerWrapper> in IndexPageGrid


pages/index.mdx
```diff
@@ -0,0 +1,224 @@
...
+            />
+          </StyledLink>
+        </CardGrid>
+        <ExternalCTA
+          icon={developerCommunityIcon}
+          title="Developer Support"
+          description="Interact with the developer community and Dwolla developer support to find answers to your technical questions."
+          href="https://discuss.dwolla.com/"
+        />
+        <H3>SDKs &amp; Tools</H3>
```
Finholt, line 204: For the h3's on this page, could we add a bit more space above them and remove a bit of the space below them to line up with the designs a bit better? The sections feel like the blend together a bit too much right now

pages/index.mdx
```diff
@@ -0,0 +1,224 @@
...
+            target="_blank"
+          >
+            <Card
+              icon={apiReferenceIcon}
+              link
+              topic="API Reference"
+              description="View details for all of Dwolla’s API endpoints with a complete reference"
+            />
+          </StyledLink>
+        </CardGrid>
```
Finholt: I think we'll have to stack these vertically at mobile breakpoints somehow. They're pushing off the screen right now

pages/index.mdx
ShreyaThapa: Good call! I can add that!

pages/index.mdx
ShreyaThapa: D'oh! Just now realizing that I didn't even think about mobile 🤦‍♀️

pages/index.mdx
```diff
@@ -47,12 +48,13 @@ export const BannerWrapper = styled.div`
...
+    margin: unset;
+    margin: 60px 0px 17px 0px;
+  }
 `;
 
-export const CardGrid = styled.div`
-  display: grid;
-  grid-template-columns: 1fr 1fr 1fr;
-  column-gap: 30px;
+export const CardGridWrapper = styled.div`
```
ShreyaThapa, line 57: This will make the cards stack on top of each other like the Guides/Concepts cards.

pages/index.mdx
ShreyaThapa: This [commit](https://github.com/Dwolla/dev-portal/pull/73/commits/9a751192eb07d682ba6d686a6a5126ab6017d0e0) should fix that! I ended up using the good old CardGrid.

pages/index.mdx
```diff
@@ -47,12 +48,13 @@ export const BannerWrapper = styled.div`
 
 export const ContentContainer = styled.div`
   padding: 60px;
+  > h3 {
```
ShreyaThapa, line 51: This may be a hacky workaround. 

I tried ditching importing Typography.H3 completely, but the margins weren't working with custom styles for me for some reason 🤔 I'll put a pin on this one to revisit.

Finholt (approved):
Approving because I don't want to hold this up any further ahead of the demo(s). I think we'll need a follow-up PR that adds a variant of the card grid component that applies `justify-content: center`.

All in all, this page is looking really good!

ShreyaThapa:
Yeah, that sounds good! I'll merge this in and follow up!

PR #80, opened 2020-06-02
WEB-689: Implementing Partnerships Page

Finholt:


components/base/Card.tsx
```diff
@@ -15,39 +15,58 @@ import { ROBOTO, POPPINS } from "../typography";
...
+    <Link href={link.href} passHref>
+      <StyledLink
+        target={link.external && "_blank"}
+        rel={link.external && "noopener noreferrer"}
+      >
+        <CoreCard link={link} {...rest} />
+      </StyledLink>
+    </Link>
+  ) : (
+    <CoreCard link={link} {...rest} />
```
Finholt, line 68: Adding an `href` to the `link` prop will allow the linking to be handled within the component rather than outside of it, but it is not required. This should also leave the door open should we ever want clicking a card to trigger an event rather than linking somewhere

components/base/CardGrid.tsx
```diff
@@ -26,16 +30,21 @@ const CardWrapper = styled.div`
     text-decoration: none;
   }
 `;
+
 type Props = {
   children: JSX.Element[];
   variant?: "default" | "center";
+  maxColumns?: 1 | 2 | 3 | 4;
```
Finholt, line 37: We can now specify a maximum number of columns to use. This was useful for the partnerships page because the "On this Page" reduced the available width for the other content, so the four column layout looked too tight, even on larger displays

package.json
```diff
@@ -62,7 +62,7 @@
     "next-images": "1.3.0",
     "next-mdx-enhanced": "^2.4.0",
     "next-plugin-yaml": "^1.0.1",
-    "prettier": "^1.19.1",
+    "prettier": "2.0.5",
```
Finholt, line 65: There was an issue with the version of prettier that we were using with using next's `Link` component in MDX files: https://github.com/prettier/prettier/issues/7176

package.json
```diff
@@ -4,7 +4,7 @@
   "private": true,
   "scripts": {
     "_eslint": "eslint . --ext .js --ext .jsx --ext .ts --ext .tsx",
-    "_prettier": "prettier .babelrc *.{js,json} \"./!(*node_modules|.next|coverage|.storybook-out|out)/*.{js,jsx,ts,tsx,css,scss}\" \"./!(*node_modules|.next|coverage|.storybook-out|out)/**/*.{js,jsx,ts,tsx,css,scss}\" \"./!(*node_modules|.next|coverage|.storybook-out|out)/**/**/*.{js,jsx,ts,tsx,css,scss}\"",
+    "_prettier": "prettier *.{js,json} \"./!(*node_modules|.next|coverage|.storybook-out|out)/*.{js,jsx,ts,tsx,css,scss}\" \"./!(*node_modules|.next|coverage|.storybook-out|out)/**/*.{js,jsx,ts,tsx,css,scss}\" \"./!(*node_modules|.next|coverage|.storybook-out|out)/**/**/*.{js,jsx,ts,tsx,css,scss}\"",
```
Finholt, line 7: The new version of prettier error-ed because we don't have a `.babelrc` file

pages/index.mdx
```diff
@@ -201,11 +200,25 @@ export default function HomePage() {
...
         </LogoLinksWrapper>
-        <H3>Integrate With Other Platforms</H3>
+        <H3>
+          <Link href="/integrations" passHref>
+            <StyledLink>
+              Integrate With Other Platforms
+              <CaretRightIcon height={11} />
+            </StyledLink>
+          </Link>
+        </H3>
```
Finholt, line 221: Linking to partnerships page here as mentioned in Slack

pages/integrations/index.mdx
```diff
@@ -0,0 +1,65 @@
...
+## Applying to the Integration Partner Program
+
+Before creating an integration application, you must apply to become an integration partner. Our partnerships team will guide you through the process of developing and publishing an integration to Dwolla’s Integration Directory. To ensure a seamless and secure integration, all integration partners will follow a set of guidelines and requirements prior to becoming listed as an approved partner.
+
+<Button
+  text="Apply to be an Integration Partner"
+  size="large"
+  variant="primary"
+  link={{
+    href: "https://www.dwolla.com/integrations/#cta-section",
```
sausman: I think this may have changed to https://www.dwolla.com/integrations/#cta-form-0

sausman (changes requested):

pages/integrations/index.mdx
Finholt: Good catch, Stephen! Should be fixed with my latest commit

sausman (approved):

PR #86, opened 2020-06-04
WEB-699: Adding Google Analytics + Tag Manager

Finholt:
I'm confident that GTM is working properly, but I might need help verifying that GA is setup and tracking page views properly

### DIFF
--- a/pages/_app.tsx
+++ b/pages/_app.tsx
@@ -1,6 +1,7 @@
-import React, { useState } from "react";
+import React, { useState, useEffect } from "react";
 import App, { createUrl } from "next/app";
 import useSWR from "swr";
+import TagManager from "react-gtm-module";
 import Layout from "../components/layout/Layout";
 import Pages from "../modules/pages";
 import { AnchorsProvider } from "../components/util/Anchors";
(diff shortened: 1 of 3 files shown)


pages/_app.tsx
```diff
@@ -13,6 +15,31 @@ import { ReactComponent as ConceptsIcon } from "../assets/images/component-icons
 import { ReactComponent as SdksToolsIcon } from "../assets/images/component-icons/side-nav/sdks-tools-nav-icon.svg";
 import "react-tippy/dist/tippy.css";
 
+const GoogleTagsAndAnalytics = ({ router }: any) => {
+  useEffect(() => {
+    if (typeof window !== "undefined") {
+      TagManager.initialize({ gtmId: "GTM-WRT5LD6" });
+    }
+    ReactGA.initialize("UA-30404064-1");
```
sddaniels: Is Google Analytics not set up through GTM?

sddaniels:
Let's try to find the resident/remaining GA expert tomorrow

pages/_app.tsx
spencerhunter: I was thinking maybe yes, but ended up both to the ticket. My fault. Agreed though, Let's see if it can all be managed through GTM!

spencerhunter: @Finholt , I'm working on adding google analytics so it's managed through GTM now.

PR #91, opened 2020-06-10
DR-582: Track client-side page views, language selections

sausman:


modules/ga.ts
```diff
@@ -0,0 +1,7 @@
+export default function ga(...args) {
+  if ("ga" in window) {
+    window.ga(...args);
+  } else {
+    console.log("window.ga not available.", JSON.stringify(args, null, 2));
```
sausman, line 5: We can remove this if we want, but it could be nice in development.

Finholt (approved):

PR #93, opened 2020-06-10
DR-575: Addressing Browser Compatibility Issues

Finholt:


components/base/Banner.tsx
```diff
@@ -12,6 +12,7 @@ const StyledBanner = styled.div`
   padding: 40px;
   display: flex;
   flex-direction: column;
+  justify-content: space-around; /* Supporting Legacy Edge */
```
Finholt, line 15: `space-evenly` is not available in Legacy (non-Chromium) versions of Edge

components/base/select/Select.tsx
```diff
@@ -27,7 +27,7 @@ const DefaultDropdownIndicator = (props: any) => {
 const CodeDropdownIndicator = (props: any) => {
   return (
     <components.DropdownIndicator {...props}>
-      <DropdownArrowDouble width={7} />
+      <DropdownArrowDouble width={7} height={12} />
```
Finholt, line 30: Safari: The select box used in the Code Examples component is rendering incorrectly

components/layout/Layout.tsx
```diff
@@ -114,6 +120,18 @@ export default function Layout({
...
+      enableBodyScroll(document.querySelector("#body-scroll-lock-side-nav"));
+      if (sidebarToggled) {
+        disableBodyScroll(document.querySelector("#body-scroll-lock-side-nav"));
+      }
+    }
+    return () => {
+      clearAllBodyScrollLocks();
+    };
+  }, [sidebarToggled]);
+
```
Finholt, line 135: Disabling the ability to scroll the content behind the open side nav on mobile. The driving force was an issue I noticed on my phone where the content was briefly visible while scrolling when the url bar went away

components/layout/SideNav.tsx
```diff
@@ -88,13 +88,17 @@ const sortByWeight = (items) => sortBy(items, (i) => i?.weight);
 
 // components
 const Container = styled.div`
-  height: 100%;
+  flex-grow: 1;
+  display: flex;
+  flex-direction: column;
+  overflow: hidden;
```
Finholt, line 94: Using height with a flex-grow parent doesn't work on safari, so I'm just chaining flexboxes all the way down to the element that needs to be able to scroll

Finholt (commented):
This PR addresses the issues in the comment I left [here](https://dwolla.atlassian.net/browse/DR-575?focusedCommentId=38831)

components/base/Banner.tsx
Finholt: Okay, cool, so this didn't work how it should've in an actual Edge browser 🙄 . I don't have time to fix it before I have to leave, but if everything else looks okay, I say merge this in for me and I'll take another crack at this particular item when I get back!

components/layout/Layout.tsx
ShreyaThapa: Nice! Thanks for picking this up! I noticed it too, and created a ticket this morning, but I should have looked here first! 😅

components/base/Banner.tsx
ShreyaThapa: Sounds good @Finholt! Everything else looks grade A!

ShreyaThapa (approved):

sausman (approved):

PR #94, opened 2020-06-11
Wrap table cell texts on larger screens

ShreyaThapa:
The resolve I went for with tables was to have the text wrap around on larger screens so the table would not need to be scrollable unnecessarily.
On smaller screens, the table is horizontally scrollable with each row displaying on a single line.


components/base/Table.tsx
```diff
@@ -35,6 +41,10 @@ const StyledTable = styled.div`
     }
     :nth-last-of-type(1) {
       border-right: 1px solid ${GREY_2};
+      /* This ensures the last column width doesn't shrink to a point where the text wraps around each word */
```
ShreyaThapa, line 44: Added a min-width on the last column to prevent this:
Screen Shot 2020-06-11 at 2 36 42 PM

sausman (approved):

PR #97, opened 2020-06-12
Added link prop to inlineCTA component

ShreyaThapa:


components/base/InlineCTA.tsx
```diff
@@ -54,17 +58,22 @@ const ArrowStyle = styled.div`
 type Props = {
   icon: string;
   text: string;
+  href: string;
```
Finholt, line 61: Will we ever use this component for external links (that would open in a new tab)? For other link components like the Card, I've been using `link?: { href?: string; external?: boolean };` as a prop to allow for that

It's totally fine if we're not planning on using this particular component like that, but I wanted to make sure we considered it!

ShreyaThapa: Good question! We have an external CTA component that has a different design (perhaps to distinguish it between external and internal? not sure 🤷 ) but I wouldn't be against accounting for both external and internal linking via the Inline CTA.

Finholt: Oh that's a good point! I forgot we had another component for that specifically. Thanks for reminding me 🙂

Finholt (approved):

components/base/InlineCTA.tsx
ShreyaThapa: Thanks for the APPROVED. As an aside, if in the future we do decide to use this CTA for an external link, we can always add that option!

PR #100, opened 2020-06-16
DR-580: Add spacing between components, remove <br> tags

sausman:


components/partial/MDXStyleWrapper.tsx
```diff
@@ -99,6 +99,11 @@ const MDXStyleWrapper = ({ children }: { children: any }) => (
           ${codeStyles.styles}
         }
       }
+
+      > div {
```
sausman, line 103: I ended up just adding some default margin to root `div` elements within MDXWrapper.

components/base/Typography.tsx
```diff
@@ -28,7 +28,7 @@ export const headingStyles = {
   h1: css`
     font-size: 42px;
     line-height: 52px;
-    margin-top: 0;
+    margin-top: 26px;
```
sausman, line 31: Added some top margin to this as well since I think we were under the assumption that `h1`s would only appear at the top of the page.

pages/guides/personal-verified-customer/handle-verification-statuses.mdx
```diff
@@ -108,7 +107,7 @@ var requestBody = {
   ssn: "202-99-1516",
 };
 
-appToken.post(customerUrl, requestBody).then(function(res) {
+appToken.post(customerUrl, requestBody).then(function (res) {
```
ShreyaThapa, line 110: Everything looks good 👍 

Wondering if this is prettier that's adding a space here? 
Also noticed that prettier had cleaned up some of the Javascript code examples and changed them from 
> appToken
    .post(customerUrl, requestBody)
    .then(function(res){...}

to
> appToken.post(customerUrl, requestBody).then(function(res) {

I guess we could manually exclude parts of the file from applying prettier, to format the code examples the way we want, but I guess this works too. Thoughts?

ShreyaThapa (approved):

components/base/Typography.tsx
Finholt: Could you elaborate on this? If `h1`s are (generally) reserved for the top of the page, wouldn't we want them to not have any top margin so they align with the "On This Page" better?

sausman: Yeah we have some pages with `h1`s in the middle of the page, but if we want we could make the top margin 0 for the first `h1` on the page to make it line up with the "On This Page" stuff. I kinda like it with the additional spacing on the top to match the whitespace below the heading, but I could go either way.

Finholt: Fair enough! Thanks for considering it; I'll defer to your judgement

Finholt (approved):

PR #101, opened 2020-06-17
Updated R08 return code actions table

ShreyaThapa:


modules/ga.ts
```diff
@@ -2,6 +2,7 @@ export default function ga(...args) {
   if ("ga" in window) {
     window.ga(...args);
   } else {
+    // eslint-disable-next-line no-console
```
ShreyaThapa, line 5: The **console** statement was causing a `warning  Unexpected console statement  no-console` when executing `yarn checks` so added this quick fix here.

sausman (approved):

PR #102, opened 2020-06-17
DR-575: Refactoring Banner to be more consistent across browsers

Finholt:
Following up on one issue that was outstanding following my PR from last week


Finholt:
Primary difference is shifting from setting a fixed height for the component with spacing distributed evenly to height that is based on the content with specified margins between the different pieces. This let me take out the `justify-content` that legacy Edge was having trouble with

ShreyaThapa (approved):
Looking good! 👍

PR #103, opened 2020-06-17
DR-584: Add anchor links to headings

sausman:


components/partial/MDXStyleWrapper.tsx
```diff
@@ -33,6 +33,10 @@ const MDXStyleWrapper = ({ children }: { children: any }) => (
 
       > h1 {
         ${headingStyles.h1.styles}
+
+        &:first-of-type > a {
```
sausman, line 38: This is to prevent the first `h1`'s anchor link from showing.

layouts/index.tsx
```diff
@@ -37,6 +42,9 @@ const MDX_COMPONENTS = {
   CardGrid,
   Card,
   code: InlineCode,
+  h1: AnchoredH1,
```
sausman, line 58: For now I only have it providing anchor links for `h1..h3` to match the listing in On This Page, but we can always add more if we want.

Finholt (commented):
Since I've been out for a bit, please forgive me if these points have already been addressed in conversations that I've missed
- My understanding is that the icon should only appear when hovering over the `h` tag, and it shouldn't be orange or underlined. Our default link styles + always being visible looks off to me
- Weren't we supposed to get an icon from design that we could use instead of a `#`?
- Having the click just scroll the page to the section that is already on the page seems kinda pointless 🤷 . I guess that's what Stripe does, so I'm not going to push this point, but I would think having the click copy the link to that section to the clipboard would add more value

ShreyaThapa (approved):

PR #104, opened 2020-06-18
DR-572: Adding CardGrid variant for homepage

Finholt:


ShreyaThapa (approved):
Thanks Connor! The 2col cards layout looks much better now than the awkward centered third card on medium screens. 😂

PR #106, opened 2020-06-23
Dr 565/accessibility

ShreyaThapa:
- [x] FilterTabs

- [x] Collapsible

- [ ] SideNav 

Pushing these up for review, while I work on the SideNav.

### DIFF
--- a/components/base/Collapsible.tsx
+++ b/components/base/Collapsible.tsx
@@ -1,4 +1,6 @@
+/* eslint-disable jsx-a11y/no-noninteractive-tabindex */
 /** @jsx jsx */
+import { useState } from "react";
 import ReactCollapsible from "react-collapsible";
 import styled from "@emotion/styled";
 import { css, jsx } from "@emotion/core";
@@ -6,10 +8,16 @@ import plusIcon from "../../assets/images/component-icons/plus-circle-icon.svg";
 import minusIcon from "../../assets/images/component-icons/minus-circle-icon.svg";
 import { GREY_2, WHITE_PRIMARY, PARAGRAPH_TEXT } from "../colors";
 import { POPPINS, ROBOTO } from "../typography";
-import { BOX_SHADOW_7 } from "../shadowDepths";
+import { BOX_SHADOW_7, BOX_SHADOW_4 } from "../shadowDepths";
 
 // Styles
+
 const CollapsibleWrapper = styled.div`
+  :focus {
+    outline: none;
+    box-shadow: ${BOX_SHADOW_7};
+  }
+
   /* Main container element for react-collapsible https://www.npmjs.com/package/react-collapsible#collapsible */
   .Collapsible {
     box-sizing: border-box;
@@ -21,7 +29,7 @@ const CollapsibleWrapper = styled.div`
     font-size: 18px;
     line-height: 27px;
     :hover {
-      box-shadow: ${BOX_SHADOW_7};
+      box-shadow: ${BOX_SHADOW_4};
     }
   }
 
@@ -36,9 +44,6 @@ const CollapsibleWrapper = styled.div`
     font-size: 16px;
     letter-spacing: 0;
     line-height: 28px;
-  }
-  span:focus {
-    outline: none;
   }
 `;
 
@@ -60,6 +65,8 @@ type Props = { triggerText: string; children: any };
 
 // Collapsible component
 function Collapsible({ triggerText, children }: Props) {
+  const [active, setActive] = useState(false);
+
   // Main content when collapsible is closed
   const closedTrigger = (
     <StyledTrigger>
@@ -78,15 +85,33 @@ function Collapsible({ triggerText, children }: Props) {
 
   const transitionTime = 200;
 
+  // Sets active state on collapsible
+  function handleActive() {
+    if (!active) {
+      setActive(true);
+    } else {
+      setActive(false);
+    }
+  }
+
+  // Handle "Enter" key to open/close collapsible
+  function handleKeyDown(e) {
+    if (e.keyCode === 13) handleActive();
+  }
+
   return (
-    <CollapsibleWrapper>
+    <CollapsibleWrapper
+      tabIndex={0}
+      onKeyDown={(keyPress) => handleKeyDown(keyPress)}
+      onClick={() => handleActive()}
+    >
       <ReactCollapsible
         trigger={closedTrigger}
         triggerWhenOpen={openTrigger}
         transitionTime={transitionTime}
-        tabIndex={0}
+        open={active}
       >
-        {children}
+        <div tabIndex={active ? 0 : -1}>{children}</div>
       </ReactCollapsible>
     </CollapsibleWrapper>
   );
--- a/components/base/Collapsible.tsx
+++ b/components/base/Collapsible.tsx
@@ -4,6 +4,7 @@ import { useState } from "react";
 import ReactCollapsible from "react-collapsible";
 import styled from "@emotion/styled";
 import { css, jsx } from "@emotion/core";
+import classnames from "classnames";
 import plusIcon from "../../assets/images/component-icons/plus-circle-icon.svg";
 import minusIcon from "../../assets/images/component-icons/minus-circle-icon.svg";
 import { GREY_2, WHITE_PRIMARY, PARAGRAPH_TEXT } from "../colors";
@@ -32,19 +33,6 @@ const CollapsibleWrapper = styled.div`
       box-shadow: ${BOX_SHADOW_4};
     }
   }
-
-  /* Content element for react-collapsible https://www.npmjs.com/package/react-collapsible#collapsible__contentinner */
-  .Collapsible__contentInner {
-    margin: 0px 70px 25px 30px;
-    * {
-      margin: unset;
-    }
-    color: ${PARAGRAPH_TEXT};
-    font-family: ${ROBOTO};
-    font-size: 16px;
-    letter-spacing: 0;
-    line-height: 28px;
-  }
 `;
 
 const StyledTrigger = styled.div`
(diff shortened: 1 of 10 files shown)


components/base/Collapsible.tsx
```diff
@@ -1,15 +1,23 @@
...
 import minusIcon from "../../assets/images/component-icons/minus-circle-icon.svg";
 import { GREY_2, WHITE_PRIMARY, PARAGRAPH_TEXT } from "../colors";
 import { POPPINS, ROBOTO } from "../typography";
-import { BOX_SHADOW_7 } from "../shadowDepths";
+import { BOX_SHADOW_7, BOX_SHADOW_4 } from "../shadowDepths";
 
 // Styles
+
 const CollapsibleWrapper = styled.div`
+  :focus {
```
ShreyaThapa, line 17: Focus style for the Collapsible container.

components/base/Collapsible.tsx
```diff
@@ -21,7 +29,7 @@ const CollapsibleWrapper = styled.div`
     font-size: 18px;
     line-height: 27px;
     :hover {
-      box-shadow: ${BOX_SHADOW_7};
+      box-shadow: ${BOX_SHADOW_4};
```
ShreyaThapa, line 33: Changed the Hover state to a different Box-shadow to somehow differentiate it from a focus state. Still not sure this works well. Open to any ideas and thoughts!

components/base/Collapsible.tsx
```diff
@@ -60,6 +65,8 @@ type Props = { triggerText: string; children: any };
 
 // Collapsible component
 function Collapsible({ triggerText, children }: Props) {
+  const [active, setActive] = useState(false);
```
ShreyaThapa, line 73: `true` `active` state refers to the Collapsible being open.
This helps manage keyboard clicks.

components/base/Collapsible.tsx
```diff
@@ -78,15 +85,33 @@ function Collapsible({ triggerText, children }: Props) {
...
+  function handleActive() {
+    if (!active) {
+      setActive(true);
+    } else {
+      setActive(false);
+    }
+  }
+
+  // Handle "Enter" key to open/close collapsible
+  function handleKeyDown(e) {
```
ShreyaThapa, line 99: Pressing enter when the the collapsible is **open**, will **close** it, and vice versa.

components/base/Collapsible.tsx
```diff
@@ -78,15 +85,33 @@ function Collapsible({ triggerText, children }: Props) {
...
+    if (e.keyCode === 13) handleActive();
+  }
+
   return (
-    <CollapsibleWrapper>
+    <CollapsibleWrapper
+      tabIndex={0}
```
ShreyaThapa, line 105: Adding focus to the container rather than using the `tabIndex` prop for ReactCollapsible for styling issues.

components/base/Collapsible.tsx
```diff
@@ -78,15 +85,33 @@ function Collapsible({ triggerText, children }: Props) {
...
+      onKeyDown={(keyPress) => handleKeyDown(keyPress)}
+      onClick={() => handleActive()}
+    >
       <ReactCollapsible
         trigger={closedTrigger}
         triggerWhenOpen={openTrigger}
         transitionTime={transitionTime}
-        tabIndex={0}
+        open={active}
```
ShreyaThapa, line 113: This sets the open/close state of the Collapsible. Starts out closed, and changes `active` state based on clicks and keypresses.

components/base/Collapsible.tsx
```diff
@@ -78,15 +85,33 @@ function Collapsible({ triggerText, children }: Props) {
...
       >
-        {children}
+        <div tabIndex={active ? 0 : -1}>{children}</div>
```
ShreyaThapa: `active` state is also used here to render the inner content tabbable when the Collapsible is open, and untabbable when the Collapsible is closed.

components/base/FilterTabs.tsx
```diff
@@ -8,7 +8,10 @@ import classnames from "../../modules/classnames";
 import { slugify } from "../../modules/helpers";
 
 // Styles
-const TabsContainerStyle = styled.div`
+const TabsContainerStyle = styled.ul`
```
ShreyaThapa, line 11: Changing the `<div>`s to `<ul>` and `<li>`s enables the tabs to be tabbable.

components/base/FilterTabs.tsx
```diff
@@ -69,11 +75,13 @@ export default function FilterTabs({ tabs, filter, setFilter }: Props) {
       <TabsContainerStyle>
         {formattedTabs.map((tab) => (
           <TabStyle
+            tabIndex={0}
             className={classnames("tab", {
               "is-active":
                 JSON.stringify(formattedFilter) === JSON.stringify(tab),
             })}
             key={tab.value}
+            onFocus={() => setFilter(tab)}
```
ShreyaThapa: Instead of having different Hover/Filter styles, I ended up changing the Filter on focus as well. In other words, focus will render filtered content based on tab change.

components/base/Collapsible.tsx
```diff
@@ -78,15 +85,33 @@ function Collapsible({ triggerText, children }: Props) {
...
+    if (!active) {
+      setActive(true);
+    } else {
+      setActive(false);
+    }
+  }
+
+  // Handle "Enter" key to open/close collapsible
+  function handleKeyDown(e) {
+    if (e.keyCode === 13) handleActive();
```
Finholt: `e.key === "Enter"` could make this a bit more readable

components/base/Collapsible.tsx
Finholt: I dig it! It's subtle, but I think it works!

components/base/Collapsible.tsx
```diff
@@ -78,15 +85,33 @@ function Collapsible({ triggerText, children }: Props) {
 
   const transitionTime = 200;
 
+  // Sets active state on collapsible
+  function handleActive() {
+    if (!active) {
+      setActive(true);
+    } else {
+      setActive(false);
+    }
```
Finholt: This could just be `setActive(!active)`

components/base/Collapsible.tsx
Finholt: Is there a way to make it so that the links are tabbable (which I assume was the driving force for this change) without making `<p>` tags tabbable? Maybe `tabIndex={active ? undefined : -1}` would do that? I honestly don't know

components/base/FilterTabs.tsx
Finholt: Hmmm, I really liked this when you were describing it, but now that I try it out, I'm not sure this is the best solution. If someone is navigating by keyboard only (as would be the case with a screen reader), they would be unable to get to the cards without inadvertently changing the filter to the last option

components/base/Collapsible.tsx
ShreyaThapa: Yeah, I originally made the `<p>` tags tabbable so that screen-readers are able to read the content. I'm actually not quite sure how people use screen readers, but I was using the ChromeVox extension for this. On the other hand, the links are tabbable regardless of making the content tabbable or not.

components/base/FilterTabs.tsx
ShreyaThapa: Oh wow! Great catch! Can't believe I missed that. So the best solution is probably something similar to how keyboard selection works on the `<select>` component. Thanks Connor!

components/base/Collapsible.tsx
ShreyaThapa: I agree. I was unsure what the best practice was.

components/base/Collapsible.tsx
ShreyaThapa: Yo! Good point! I was really struggling with the eradication of the redundancy here with these nested if statements. Thanks for the free point ;)

components/base/Collapsible.tsx
Finholt: Interesting! Mac's default screenreader VoiceOver doesn't have trouble reading them either way, but if ChromeVox is, maybe we could just hide the border that shows up on the focus state right now instead?

ShreyaThapa: If it works for Mac, it might work for other screen readers too? 🤷‍♀️ Does having the inner content be tabbable mess with the Mac Screen reader? If so, I can remove the focus altogether! Otherwise, just removing the focus outline works too!

ShreyaThapa: Ah nvm! You're right, it works in ChromeVox too. I'll remove the tabIndex from the content. 👍

components/base/Collapsible.tsx
```diff
@@ -111,7 +115,13 @@ function Collapsible({ triggerText, children }: Props) {
...
         open={active}
       >
-        <div tabIndex={active ? 0 : -1}>{children}</div>
+        <InnerContentWrapper
+          className={classnames({
+            visuallyhidden: !active,
+          })}
+        >
+          {children}
+        </InnerContentWrapper>
```
ShreyaThapa, line 121: Collapsible innerContent is now visually hidden when not opened. This disables the links/content from being tabbable when the Collapsible is closed.

components/base/FilterTabs.tsx
```diff
@@ -81,10 +87,10 @@ export default function FilterTabs({ tabs, filter, setFilter }: Props) {
                 JSON.stringify(formattedFilter) === JSON.stringify(tab),
             })}
             key={tab.value}
-            onFocus={() => setFilter(tab)}
             onClick={() => {
               setFilter(tab);
             }}
+            onKeyDown={(keyPress) => onKeydown(keyPress, tab)}
```
ShreyaThapa, line 93: Filters are only changed if the User clicks or presses enter on the keyboard.

components/base/Collapsible.tsx
ShreyaThapa: This should fix it - https://github.com/Dwolla/dev-portal/pull/106/commits/51f1819764827bc7b1a31862d3b378b6eb2f8f33

components/base/FilterTabs.tsx
ShreyaThapa: This should fix it - https://github.com/Dwolla/dev-portal/pull/106/commits/51f1819764827bc7b1a31862d3b378b6eb2f8f33

components/base/Collapsible.tsx
```diff
@@ -60,6 +48,23 @@ const ImageStyles = css`
...
+  }
+  color: ${PARAGRAPH_TEXT};
+  font-family: ${ROBOTO};
+  font-size: 16px;
+  letter-spacing: 0;
+  line-height: 28px;
+
+  &.visuallyhidden {
+    transition-delay: 200ms;
+    visibility: hidden;
```
ShreyaThapa, line 64: Added a transition to the make the inner content visible while the Collapsible is **closing**.

Finholt (approved):

PR #112, opened 2020-06-29
DEV-240: Fixing iOS Sidebar Scroll Issue + Other Tweaks

Finholt:


components/layout/TopBar.tsx
```diff
@@ -184,7 +184,11 @@ export default function TopBar({
 
   return (
     <Container>
-      <StyledLogo src={dwollaDevLogo} alt="" />
+      <Link href="/">
+        <a>
+          <StyledLogo src={dwollaDevLogo} alt="" />
+        </a>
+      </Link>
```
Finholt: Realized the link was added only to the side nav, so added it to the (mobile) top bar and footer as well

components/partial/MDXStyleWrapper.tsx
```diff
@@ -100,7 +100,7 @@ const MDXStyleWrapper = ({ children }: { children: any }) => (
         }
       }
 
-      > div {
+      > div:not(.mdx-style-override) {
         margin-top: 29px;
         margin-bottom: 29px;
       }
```
Finholt, line 106: These margins were to add whitespace between adjacent components, but had the side effect of adding that margin to the banner components on the guide/concept index pages as well. This removes that gap again

components/layout/SideNav.tsx
```diff
@@ -99,11 +100,14 @@ const Slide = styled.div`
...
+  position: absolute;
+  top: 0;
+  bottom: 0;
+  left: 0;
+  transition: left ease 0.4s;
 
   &.slidIn {
-    transform: translate3d(-50%, 0, 0);
+    left: -100%;
   }
```
Finholt: For whatever reason, Safari seems to have an issue using `transition` and `transform` together. I still don't know the root reason why, but it looks like this has been an issue for at least 5 years based on when this exact same issue was brought up in [this StackOverflow question](https://stackoverflow.com/questions/28288710/cant-scroll-inside-a-div-after-applying-webkit-transform-in-safari) 🤦

Finholt: Of course this fix only worked for desktop Safari and iOS is still broken 🤦 🤦

Finholt:
Okay, so I couldn't figure out exactly what was causing the issue, but I took a crack at refactoring it to be a little simpler and whatever the root cause of the issue looks to have been removed as a result. The difference is that instead of rendering two slide panes and then sliding them back and forth on a "track" of sorts, we now only render one slide pane whose content changes and conditionally re-renders based on the section (the necessary animation plays based on this as well)

This had the added benefit of simplifying some of our accessibility stuff because there's no longer an entire nav pane hanging out outside of view that we have to control tabbing for

components/layout/SideNav.tsx
```diff
@@ -341,11 +343,6 @@ const MobileWrap = styled.div`
   @media (${breakUp("lg")}) {
     display: none;
   }
-
-  /* Hides MobileWrap contents when Sidebar is toggled and a Section is active. Not tabbable when not in view. */
-  &.visuallyhidden {
-    visibility: hidden;
-  }
```
Finholt, line 348: Because there is now only one "SlidePane", there is no longer a duplicate MobileWrap that needs to be hidden

components/partial/MDXStyleWrapper.tsx
ShreyaThapa: Oops! Good catch!

ShreyaThapa (approved):

PR #119, opened 2020-07-09
DEV-417: Convert External Links from <Link> to <a>

ShreyaThapa:


components/base/Button.tsx
```diff
@@ -127,14 +126,13 @@ const StyledLink = styled.a`
...
-      {text}
-    </StyledLink>
-  </Link>
+  <StyledLink
+    href={link.href}
+    target={link.external && "_blank"}
+    className={`${variant} ${size}`}
+  >
+    {text}
+  </StyledLink>
```
ShreyaThapa, line 135: Removed <Link> from buttons since they're all external links. If in the future we need to add Buttons that link internally, we can add a conditional wrapper.

components/layout/TopBar.tsx
```diff
@@ -104,10 +104,21 @@ export type TopBarLinkProps = {
   active?: boolean;
 };
 
+const ConditionalLinkWrapper = ({ external, linkWrapper, children }) =>
+  external ? children : linkWrapper(children);
+
```
ShreyaThapa, line 109: Wrapping TopBarLink item in a <Link> based on the `external` prop.

sausman (approved):

PR #126, opened 2020-07-17
DEV-261: Refactor active anchor detection

sausman:


components/util/Anchors.tsx
```diff
@@ -73,9 +87,62 @@ export function AnchorsSetter(props) {
...
 
-    return () => setAnchors([]);
+    return () => {
+      setAnchors([]);
+      setActiveAnchor(null);
+    };
   }, [router?.pathname]);
 
+  useLayoutEffect(() => {
+    let raf: number;
```
sausman, line 99: `raf` will be used to store a reference (number) to a `requestAnimationFrame` which we can use to cancel it later if we need to.

components/util/Anchors.tsx
```diff
@@ -73,9 +87,62 @@ export function AnchorsSetter(props) {
...
+    const anchorEls = anchors.map((a: Anchor) => document.getElementById(a.id));
+
+    const handleScroll = () => {
+      raf = requestAnimationFrame(() => {
```
sausman, line 103: Here's where we set the `raf` variable. Every time `handleScroll` is called, we pass a function to `requestAnimationFrame` that we want to run every animation frame (instead of running it immediately since `handleScroll` gets called a lot).

components/util/Anchors.tsx
```diff
@@ -73,9 +87,62 @@ export function AnchorsSetter(props) {
...
+      });
+    };
+
+    window.addEventListener("scroll", handleScroll);
+    window.addEventListener("resize", handleScroll);
+
+    return () => {
+      window.removeEventListener("scroll", handleScroll);
+      window.removeEventListener("resize", handleScroll);
+      cancelAnimationFrame(raf);
```
sausman, line 130: This is the cleanup function that gets called when the component unmounts. Here we clean up our event handlers and `cancelAnimationFrame(raf)` so they don't get called anymore.

components/layout/OnThisPage.tsx
```diff
@@ -50,6 +48,7 @@ const Container = styled.div`
     box-sizing: border-box;
     border-left: 1px solid ${GREY_2};
     cursor: pointer;
+    transition: border-color 200ms ease;
```
ShreyaThapa, line 52: Nice! Love the transition. 

Could we also remove the default focus outline and add a background color like in the side-nav to the anchors in this Container? 

>  \> a {
:focus {
      outline: none;
      color: ${HEADLINE_TEXT};
      background-color: ${GREY_1};
    }
}

sausman: Good call! Updated

ShreyaThapa (approved):

PR #147, opened 2020-10-12
Added .prettierignore

ShreyaThapa:


.prettierignore
```diff
@@ -0,0 +1,9 @@
+# Ignore markdown files which have '$' signs that get excaped.
+# https://github.com/prettier/prettier/issues/6213
+
+pages/concepts/customer-types.mdx
+pages/concepts/facilitator-fee.mdx
+pages/concepts/sift.mdx
+
+pages/guides/micro-deposit-verification/index.mdx
+pages/guides/sandbox/index.mdx
```
spencerhunter, line 9: Just to clarify, moving forward we'll need to add any new file paths if those files contain a $ sign?

spencerhunter: Oh, nvm. I see the comment: https://github.com/prettier/prettier/issues/6213#issuecomment-652994522.

spencerhunter (approved):

.prettierignore
ShreyaThapa: Correct. Or when a file contains a character that is escaped, but isn't parsed properly when you run it. 

For example, check line 30 of this [file](https://github.com/Dwolla/dev-portal/blob/master/pages/guides/receive-money/create-transfer.mdx#step-4a-initiate-a-transfer).
The "_"  in "_links" is escaped like this `\_links`, but the parser works as expected and only renders `_links`.

PR #165, opened 2020-12-08
initial drop-ins docs

spencerhunter:


pages/guides/drop-ins/setup-and-configuration.mdx
```diff
@@ -0,0 +1,94 @@
...
+
+To get up and running quickly, take a look at our [drop-ins examples repo](https://github.com/Dwolla/drop-ins-examples) which is a basic Node.js app that uses [Express](https://expressjs.com/).
+
+### Include dwolla-web.js
+
+Begin the client-side implementation by including dwolla-web.js in the ‘head’ of your HTML page. The Dwolla Components library containing all of the drop-in UI components are available directly from Dwolla’s content delivery network (CDN). As we continue to evolve the components library, it will soon be available to be installed via NPM.
+
+```html
+<head>
+  <script type="text/javascript" src="cdn.dwolla.com/2/dwolla.js"></script>
```
sausman, line 23: I don't think `src="cdn.dwolla.com/..."` will work. Pretty sure it needs to be something like `src="//cdn.dwolla.com/..."`, else it will treat it as a relative path and load `https://myapp.com/cdn.dwolla.com/2/dwolla.js`.

Also, the path we have right now is `https://cdn.dwolla.com/v2/dwolla-web.js` (can be changed easily tho)

pages/guides/drop-ins/generate-client-token.mdx
```diff
@@ -0,0 +1,113 @@
...
+  _links: {
+    customer: {
+      href:
+        "https://api-sandbox.dwolla.com/customers/707177c3-bf15-4e7e-b37c-55c3898d9bf4",
+    },
+  },
+  action: "customer.update",
+};
+
+appToken.post(`/client-token`).then((res) => res.body.token); // => '4adF858jPeQ9RnojMHdqSD2KwsvmhO7Ti7cI5woOiBGCpH5krY'
```
sausman: `requestBody` isn't getting passed to `post`

Also, we can use normal quotes for `"/client-token"`. I think \`\` are only needed for string interpolation and/or es6 templates.

pages/guides/drop-ins/generate-client-token.mdx
```diff
@@ -0,0 +1,113 @@
...
+```
+
+### Send client token to front-end
+
+After obtaining a client-token, you will pass this token to your front-end. The example shown below demonstrates this process using Express.
+
+```javascript
+app.get("/update-customer", function (req, res) {
```
sausman, line 96: We might be able to get away with something like:

```javascript
// see examples above
const myClientToken = "..."

// POST /dwolla-client-token
app.post("/dwolla-client-token", (req, res) => {
  res.json({ token: myClientToken });
})
```

ShreyaThapa (approved):

pages/guides/drop-ins/setup-and-configuration.mdx
ShreyaThapa: @sausman, which one of the following should we use for this example?  
1. //cdn.dwolla.com/2/dwolla.js
or 
2. //cdn.dwolla.com/v2/dwolla-web.js

PR #167, opened 2020-12-08
Clarified what status a beneficial owner has to be in for update

kellymore:


pages/guides/business-verified-customer/add-beneficial-owners.mdx
```diff
@@ -525,7 +525,9 @@ appToken
 
 ## Remove and update a Beneficial Owner
 
-If an individual beneficial owner wants to update their information, that individual beneficial owner will first need to be removed.
+A beneficial owner's information can only be updated if their verification status is incomplete.
+
+If an individual beneficial owner needs to update their information, that individual beneficial owner will first need to be removed.
```
ShreyaThapa: Can we change this to:

>If an individual beneficial owner needs to update their information after being verified, there are two ways of accomplishing it:
> - remove the individual beneficial owner and re-add with updated information
> - contact Dwolla to place the beneficial owner into `incomplete` status again so that the owner's information can be updated

kellymore: yes!

PR #169, opened 2020-12-10
Added Numbered Pagination component

ShreyaThapa:


kellymore (approved):

components/base/NumberedPagination.tsx
```diff
@@ -0,0 +1,118 @@
...
+`;
+
+type Props = {
+  totalPages: number;
+  newerLabel: string;
+  olderLabel: string;
+};
+
+function NumberedPagination({ totalPages, newerLabel, olderLabel }: Props) {
+  const pageNumbers = [];
```
sausman, line 90: No need to change this, but out of curiosity I [found a way](https://stackoverflow.com/questions/3895478/does-javascript-have-a-method-like-range-to-generate-a-range-within-the-supp) to do this without having to mutate the array in a `for` loop:

```javascript
const pageNumbers = [...Array(totalPages).keys()];
```

ShreyaThapa: Oh this is perfect! I did feel like there had to be a better way!

I'll make this change when implementing the component! Thanks Stephen!

PR #181, opened 2021-01-20
Updated component name

kellymore:


pages/guides/drop-ins/generate-client-token.mdx
```diff
@@ -10,10 +10,16 @@ meta:
...
 # Step 2: Generate a client token
 
-Regardless of which drop-in component is being used by your application, the dwolla-web.js JavaScript library will require a unique “client-token” to be passed in on configuration. A client token contains granular permissions and is scoped to the end-user/Customer that is performing the action within your web application.
+Regardless of which drop-in component is being used by your application, the dwolla-web.js JavaScript library will require a unique “client-token” or a server side route that is used to generate a client token to be passed in on configuration. A client token contains granular permissions and is scoped to the end-user/Customer that is performing the action within your web application.
 
 Client-tokens are single-use tokens that are valid for up to 1 hour after being generated. More than one client token can be generated and valid at one time.
 
+#### Token vs. Token URL
+
+For flow components like Business Verified Customers, Personal Verified Customers, and Payins, it is recommended to use token URL. We also recommend using the token URL if you plan to use multiple drop-in components.
```
spencerhunter: Should we define what flow components are or is that obvious? 

e.g. Flow components are designed to support drop-ins that have multiple isolated functions connected together into a single user flow. Creating a server-side endpoint that can be called by the dwolla-web library when an action needs to be performed within the flow simplifies the integration experience.

spencerhunter: @misskellymore Also, maybe include a reference to the next step where they configure dwolla-web.js to use `token` and `tokenUrl`.

spencerhunter (approved):

PR #195, opened 2021-02-11
Replaced markdown styling with html tags when using AlertBar component

ShreyaThapa:
- Updated `<AlertBar>` component prop-type
- Updated `<AlertBar>` snapshot
- Replaced markdown styling with matching html tags when using `<AlertBar>`

### DIFF
--- a/app/components/base/AlertBar.tsx
+++ b/app/components/base/AlertBar.tsx
@@ -41,7 +41,7 @@ const TextWrapper = styled.div`
   line-height: 25px;
 `;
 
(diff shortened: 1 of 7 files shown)


kellymore (approved):
I would just change "This is a warning!" to "Important!". But g2g

ShreyaThapa:
> I would just change "This is a warning!" to "Important!". But g2g

Thanks for the approve, Kelly! 

Yeah, that's just a test and not public facing, so it shouldn't matter. I typed that up to match the prop option of `warning`.

PR #201, opened 2021-03-09
add beneficial owner drop-in to supported components list

spencerhunter:


pages/changelog.mdx
```diff
@@ -16,6 +16,12 @@ No upcoming changes
 
 ## Completed
 
+### 2021-03-10
+
+**ADDED**
+
+- Added Drop-in component for [adding Beneficial Owners](/concepts/drop-in-components#add-beneficial-owners). This pre-built UI component provides a low-code solution for assisting with onboarding business Verified Customers within your application. Learn more about [Drop-in components](/concepts/drop-in-components#ui-components-library) in our documentation and check out our guide on [building with Drop-in components](/guides/drop-ins#building-with-drop-in-components).
```
ShreyaThapa: Link for `adding Beneficial Owners` needs to be `/concepts/drop-in-components#create-beneficial-owners`

pages/concepts/drop-in-components.mdx
```diff
@@ -279,7 +280,48 @@ tooltip-shift,
...
+dwolla-owner-name, dwolla-owner-delete,
+dwolla-owner-header, dwolla-owners-summary-container,
+dwolla-owners-empty, dwolla-add-owners-button,
+dwolla-owner, dwolla-owner-status,
+dwolla-owner-status-verified, dwolla-owner-status-incomplete,
+dwolla-owner-status-document, dwolla-button-label-container-nb,
+dwolla-button-label-container, dwolla-link,
+dwolla-customer-tos, dwolla-customer-checkbox,
+dwolla-customer-text, dwolla-text-container,
+dwolla-input-container, dwolla-loading
```
ShreyaThapa: The # of classes appears to be similar to that of Create a Business Verified Customer. Could we give each class their own line?

ShreyaThapa (changes requested):

pages/concepts/drop-in-components.mdx
spencerhunter: I thought about that but the container was getting a bit large so I landed on this haha

pages/changelog.mdx
spencerhunter: Ah good catch. I updated that header name last minute

pages/concepts/drop-in-components.mdx
ShreyaThapa: That makes sense. 

It is kind of lengthy. However, I think it should be fine. With the `On this Page` navigation, users will probably be jumping to the section they need rather than scroll to get there.

pages/changelog.mdx
spencerhunter: Updated

pages/concepts/drop-in-components.mdx
spencerhunter: Updated

ShreyaThapa: Thanks! 🙌

ShreyaThapa (approved):
ⒶⓅⓅⓇⓄⓋⒺⒹ

PR #214, opened 2021-06-25
hero draft pr

kellymore:


ShreyaThapa (commented):
Looks good Kelly! 

Next steps, let's
- update the component name to `HeroImage`
- write a story and a test for this. 
- get a higher resolution image from UX

We can then revert the `pages/index.mdx` changes and open up the PR!

app/components/base/Hero.tsx
```diff
@@ -0,0 +1,32 @@
...
+  flex-direction: column;
+  align-items: right;
+  position: relative;
+  width: 100%;
+
+  @media (${breakDown("sm")}) {
+    display: none;
+  }
+
+  @media (${breakDown("xs")}) {
```
ShreyaThapa: We can probably get rid of this `@media` block since the breakdown for `"sm"` will account for "xs" as well.

app/components/base/__tests__/HeroImage.test.tsx
```diff
@@ -0,0 +1,13 @@
...
+test("New Hero Image", () => {
+  const tree = renderer.create(<HeroImage />).toJSON();
+
+  expect(tree).toMatchSnapshot();
+});
+
+test("Style", () => {
+  const tree = renderer.create(<StyledHero />).toJSON();
+  expect(tree).toMatchSnapshot();
+});
```
kellymore: To test attributes + styles on styled component, we'll have to use jest-styled-components, or a different testing library. What do you think?  toHaveStyle() won't work here. 

cc @ShreyaThapa

ShreyaThapa: Ah, actually the snapshot testing that we're doing with these components accounts for testing the styles as well! [Here's ](https://medium.com/styled-components/effective-testing-for-styled-components-67982eb7d42b) a good article I found that explains snapshotting!

ShreyaThapa (approved):
Approved with changes in this [commit](https://github.com/Dwolla/dev-portal/pull/214/commits/e218329d2f7cb536ee482316c8ea9b09f9448fb4).

PR #234, opened 2021-09-23
Dev 705/rtp testing

ShreyaThapa:


kellymore (changes requested):
I think the ® on RTP® may be a legal requirement. Other than that, looks good

kellymore (approved):

PR #273, opened 2021-12-08
Fix node.js.yml CI

kellymore:


.github/workflows/node.js.yml
```diff
@@ -1,29 +1,30 @@
...
 
 name: Node.js CI
 
 on:
   push:
-    branches: [main]
+    branches: [ $default-branch ]
   pull_request:
-    branches: [main]
+    branches: [ $default-branch ]
```
ShreyaThapa, line 11: Is there any significance of using `$default-branch` instead of `main`?

kellymore: Just best practices

ShreyaThapa: Got it.

ShreyaThapa (approved):

.github/workflows/node.js.yml
ShreyaThapa: Let's merge this in and test with a small PR.

PR #279, opened 2021-12-13
Update footer

kellymore:


app/components/layout/__tests__/Footer.test.tsx
```diff
@@ -31,7 +31,7 @@ test("Footer", () => {
   const FOOTER_LEGAL_COPY = {
     title: "Financial institutions play an important role in our network.",
     operatorDescription:
-      "Dwolla, Inc. is the operator of a software platform that communicates user instructions for funds transfers to our financial institution partners.",
+      "Financial Institutions are an important part of our network. All funds transfers made using the Dwolla Platform are performed by a financial institution partner, and any funds held in a Dwolla Balance are held by a financial institution partner. [Learn more about our financial institution partners](https://www.dwolla.com/legal/about-our-financial-institution-partners/#legal-content).",
     veridianDescription:
```
ShreyaThapa, line 35: Looks like we can remove the `veridianDescription`, `metaBankDescription`, and `rtpDescription` descriptions now that the new text will be replacing all. Check out the footer in https://www.dwolla.com/ for example.

ShreyaThapa (changes requested):

PR #280, opened 2021-12-13
Remove anchors

kellymore:


ShreyaThapa (commented):
I don't think we want to remove the `#` anchor signs. Instead, we want the H1s to not appear on the URL. 

For example, 
`Before:` https://developers.dwolla.com/guides/business-verified-customer#business-customer-verification
`After:` https://developers.dwolla.com/guides/business-verified-customer

A good place to start _might_ be removing the "h1" from [line 17](https://github.com/Dwolla/dev-portal/blob/main/app/components/util/Anchors.tsx#L17).

PR #293, opened 2022-01-11
Transfer lifecycle/dev 798

ShreyaThapa:
- Implement switch case to compare and set Selected Funds Flow.
- Implement concept article markdown file for Transfer Lifecycle


app/components/base/TimelineNav.tsx
```diff
@@ -120,7 +120,7 @@ const createElements = (steps, active, setActive) => {
       <StyledStep className={status} key={step}>
         <StyledStepNum
           className={status}
-          onClick={idx !== 0 && onClick}
+          onClick={idx !== 0 ? onClick : undefined}
```
ShreyaThapa, line 123: Edit conditional syntax to avoid errors.

app/components/base/TransferWorkflow.tsx
```diff
@@ -210,20 +229,109 @@ function TransferWorkflow() {
...
 
-    if (selectedSender.value === "account") {
-      //  TODO: Write if/else statements for all combinations + add JSON data
-    } else if (selectedSender.value === "vcr") {
-      if (isEqual(setSelectedFundsFlow, fundsFlowCombinations.vcr.vcr4)) {
-        setWebhookJsonData(VCRBanktoCRBank);
-      } else if (isEqual(selectedFundsFlow, fundsFlowCombinations.vcr.vcr6)) {
-        setWebhookJsonData(VCRBanktoVCRBankdata);
-      } else {
+    switch (selectedSender.value) {
```
ShreyaThapa, line 232: Using Switch/Case instead of If/Else for efficiency. Takes less time to run, and is more readable.

spencerhunter (approved):

PR #302, opened 2022-01-21
Dev 819/eslint

ShreyaThapa:
Sorry for the huge PR. Most changes are related to fixing or ignoring ESlint errors after the Node upgrade in `.nvmrc`.
Some changes fix warnings (as noted in the comments).
The second commit includes snapshot updates and can be quickly skimmed through.


.eslintrc.js
```diff
@@ -39,7 +40,13 @@ module.exports = {
...
     "react/jsx-props-no-spreading": "off",
     "react/jsx-wrap-multilines": "off", // prettier
+    "react/jsx-no-bind": "off", // typescript
+    "react/jsx-no-constructed-context-values": "off", // typescript
+    "react/no-children-prop": "off", // typescript
+    "react/no-unstable-nested-components": "off",
+    "react/no-unused-prop-types": "off", // typescript
     "react/react-in-jsx-scope": "off", // next
+    "react/require-default-props": "off", //typescript
     "react/self-closing-comp": "off", // prettier
```
ShreyaThapa, line 50: Ignoring some ESlint rules to avoid disrupting existing working code.

app/components/base/Button.tsx
```diff
@@ -142,44 +142,47 @@ type Props = {
   onButtonClick?: any;
 };
 
-const Button = ({
```
ShreyaThapa, line 145: Declaring function components as `functions` instead of `const`. 

There are a lot of these. All were auto-fixed by ESlint.

app/components/base/Card.tsx
```diff
@@ -117,55 +117,59 @@ type Props = {
...
+    </CardStyle>
+  );
+}
+
+function Card({ link, ...rest }: Props) {
+  return link && link.href ? (
     <Link href={link.href} passHref>
       <StyledLink
-        target={link.external && "_blank"}
+        target={link.external ? "_blank" : undefined}
```
ShreyaThapa, line 172: Fixes the following warning (there are a few of these): 

```
Warning: the 'url' property is deprecated. https://err.sh/zeit/next.js/url-deprecated
Warning: Received `false` for a non-boolean attribute `target`.

If you want to write it to the DOM, pass a string instead: target="false" or target={value.toString()}.

If you used to conditionally omit it with target={condition && value}, pass target={condition ? value : undefined} instead.
```

app/components/base/FundsFlowSelector.tsx
```diff
@@ -131,15 +131,15 @@ const Option = (props: any) => {
       <components.Option {...props} />
     </StyledOptionWrap>
   );
-};
+}
 
-type Option = { value: string; label: string };
+type Options = { value: string; label: string };
```
ShreyaThapa, line 136: `Option` was already being used; so renamed this const to `Options`.

app/components/layout/OnThisPage.tsx
```diff
@@ -109,8 +109,6 @@ const PageTop = styled.div`
   }
 `;
 
-type Heading = { key: string; level: number; title: string };
```
ShreyaThapa, line 112: `Heading` wasn't being used anywhere.

tarricsookdeo (approved):
Looking good!

PR #325, opened 2022-04-12
Upgrade React to v17

jcoon97:
Since Next.js is now running at v9.5.5, this PR is focused on upgrading React to v17, as Next.js will require React v17 as a peer dependency when upgrading to v10.


ShreyaThapa (approved):

ShreyaThapa (approved):
Runs and builds fine now!

PR #392, opened 2022-08-29
DEV-1020: Update business customers to include `businessName` restrictions

jcoon97:


pages/api-reference/customers/create.mdx
```diff
@@ -7,8 +7,6 @@ meta:
   title: "Create a Customer | Dwolla API Documentation"
 ---
 
-import guideIcon from "../../../assets/images/content-images/content-icons/guides-icon.svg";
```
jcoon97, line 10: Removed this import since `guideIcon` is not used anywhere else on the page. Happy to add it back, though, if requested.

ShreyaThapa (approved):

PR #413, opened 2022-09-27
DEV-1060: Update global language selector to use MUI

jcoon97:


app/components/layout/TopBar.tsx
```diff
@@ -119,15 +120,17 @@ export default function TopBar({ button, onHamburgerClick }: TopBarProps) {
...
-          selectedValue={selectedLanguage}
-          setSelectedValue={(language) => {
-            setSelectedLanguage(language);
-            ga("global language select", "language change", language.value);
-          }}
-          autoWidth
-        />
+        <FormControl sx={{ minWidth: 200 }} size="small">
+          <SelectMui
+            label="Select Language"
```
jcoon97, line 163: I used "Select Language" instead of "View as" from Figma, but I'm happy to change it back to that or something else, if we don't like this.

ShreyaThapa: Select Language fits best here.

ShreyaThapa (dismissed):

ShreyaThapa (approved):

PR #417, opened 2022-10-10
DEV-1110: Fix hydration error caused by remapping `<table>` to `<div>`

jcoon97:


app/components/base/Table.tsx
```diff
@@ -5,7 +5,7 @@ import { HEADLINE_TEXT, PARAGRAPH_TEXT, GREY_1, GREY_2 } from "../colors";
 import { breakUp } from "../breakpoints";
 
 // Styles for Table
-const StyledTable = styled.div`
```
jcoon97, line 8: When MDX (and React/Next) hydrate a `<table>` element, this line replaces it with `<div>` with some additional styling added to it. This causes a hydration error since `<thead>`, `<tbody>`, etc. are not legal children of `<div>` but rather only `<table>`.

ShreyaThapa (approved):

PR #477, opened 2023-04-12
DEV-1362: Create Funding Source drop-in component documentation

jcoon97:


pages/concepts/drop-in-components.mdx
```diff
@@ -703,6 +704,77 @@ dwolla-document-submit
   height="75%"
 />
 
+### Create a Funding Source
+
+The `dwolla-funding-source-create` component is used to display a form to collect the required information needed to create a [bank funding source](/api-reference/funding-sources). Currently, this component only allows creating a funding source that is attached to a [customer](/api-reference/customers) record. Optionally, this component can automatically [initiate micro-deposits](/api-reference/funding-sources/initiate-micro-deposits) when the funding source is created, if specified via the `initiateMicroDeposits` attribute.
```
spencerhunter, line 709: Wonder if we should add a note or CTA here to indicate that the component doesn't currently support `on-demand-authorization` to be applied?

spencerhunter (commented):
Wonder if we should also add the client-token action details [here](https://github.com/Dwolla/dev-portal/blob/DEV-1362/pages/api-reference/drop-in-components/create-client-token.mdx?plain=1#L17)?

jcoon97:
> Wonder if we should also add the client-token action details [here](https://github.com/Dwolla/dev-portal/blob/DEV-1362/pages/api-reference/drop-in-components/create-client-token.mdx?plain=1#L17)?

Addressed in [this commit](https://github.com/Dwolla/dev-portal/pull/477/commits/74317811b452479687a872355418bc7a9a815002).

pages/concepts/drop-in-components.mdx
jcoon97: Addressed in [this commit](https://github.com/Dwolla/dev-portal/pull/477/commits/2adc9fe033b531d738080564e53825dee93740c1#diff-4631ade2856e72d7be2d5698fb7fcae1170b6aa58bdd741b9f505a9f51943813R711-R720). Not married to the wording, but hopefully is a good start!

spencerhunter (dismissed):

jcoon97:
@spencerhunter, are you able to re-review when you get a chance? Previous approval went stale when I merged in 4e82d73.

ShreyaThapa (approved):

PR #500, opened 2023-05-15
DEV-1381: Side Navigation Component updates for consolidated docs

ShreyaThapa:


.prettierignore
```diff
@@ -1,9 +1,9 @@
 # Ignore markdown files which have '$' signs that get excaped.
 # https://github.com/prettier/prettier/issues/6213
 
-pages/concepts/customer-types.mdx
-pages/concepts/facilitator-fee.mdx
-pages/concepts/sift.mdx
+pages/balance/get-started/customer-types.mdx
```
spencerhunter: should we move this into the new root /docs directory? 

/docs/ - New homepage
/docs/balance - Balance docs homepage
/docs/balance/customer-types

ShreyaThapa: Thanks for taking a look at the PR and for the suggestion! 

Absolutely! I'l bundle that in with the next round of changes. 👍

PR #503, opened 2023-05-31
DEV-1383: Create page-level feedback component

jcoon97:


app/components/base/PageFeedback.tsx
```diff
@@ -0,0 +1,92 @@
...
+import {
+  type SvgIconComponent,
+  ThumbDownOffAltOutlined,
+  ThumbUpOffAltOutlined,
+} from "@mui/icons-material";
+import merge from "lodash.merge";
+import { Link } from "./Typography";
+
+const FeedbackButton = styled(Button, { name: "FeedbackButton" })(() => ({
+  color: "rgba(0, 0, 0, 0.50)",
```
ShreyaThapa: Could we add this as a variable to the `app/components/colors.ts` file? 

Here's a naming convention I've been using in other DevPortal redesign updates that might work here as well:
```
// app/components/colors.ts

// Overlay Colors
export const GREY_050 = "rgba(0, 0, 0, 0.50)";
```

app/components/base/PageFeedback.tsx
```diff
@@ -0,0 +1,92 @@
...
+const FeedbackContainer = styled(Box, { name: "FeedbackContainer" })(() => ({
+  alignItems: "center",
+  display: "flex",
+  flexDirection: "row",
+  gap: "12px",
+}));
+
+const FeedbackInnerText = styled(Typography, { name: "FeedbackInnerText" })(
+  () => ({
+    color: "rgba(106, 114, 130, 1)",
```
ShreyaThapa: Same for this one: could we add this as a variable to the `app/components/colors.ts` file?

Not sure what the naming should be for this one though 😅 😁

jcoon97:
@ShreyaThapa, I pulled the [colors out](https://github.com/Dwolla/dev-portal/pull/503/files#diff-6492e91d497c9f542a182b775d8475361a80bc6509348b5be94642c28240934bR66-R69) and prefixed them with the component name, similar to some others, like `API_STATUS_x`, etc. Hopefully that works!

ShreyaThapa (dismissed):

ShreyaThapa (approved):

PR #505, opened 2023-06-21
Dev 1385: DevPortal HomePage redesign

ShreyaThapa:


app/components/base/Help.tsx
ShreyaThapa, line 1: Using MUI's `<Menu>` and `<MenuItem>` instead of `<List*>` components. This gives the desired mouse and keyboard interaction and makes the component screen-reader friendly.

jcoon97 (approved):

PR #506, opened 2023-06-27
Move page redirect from getServerSideProps to next.config.js

jcoon97:


ShreyaThapa (approved):
Nice! Thanks for the fix!

PR #513, opened 2023-08-08
Dev 1457/homepage

ShreyaThapa:
- [x] Update homepage content copy 
- [x] Update product Icons in Cards
- [x] Move `/testing` to `/docs/balance/testing` | Category - `Get Started`


app/components/base/Card.tsx
```diff
@@ -178,12 +186,19 @@ type Props = {
   isFlex?: boolean;
   icon?: string;
   badge?: string;
-  productName?: string;
+  product?: "BALANCE" | "CONNECT";
```
jcoon97:
```suggestion
  product?: keyof typeof productDisplayText;
```
Not required, but will make it so TS tracks `productDisplayText`, rather than explicitly putting the values here

ShreyaThapa: Nice! I like this better. Thanks! :)

jcoon97 (approved):

PR #533, opened 2023-08-16
Add drop-ins and dwolla-js to Balance docs

ShreyaThapa:
I attempted to display the drop-in docs after "Testing", however, I found the doc sorting logic to be a bit too convoluted for me to make that update, so I've tabled the sorting issue for now given that this is a temporary location for the documents. We can certainly revisit the logic if we want to!

The screenshot shows how it's currently displayed in the side-nav.

Screen Shot 2023-08-16 at 5 28 00 PM


spencerhunter:
@ShreyaThapa These resources will likely be moved again before the end of the year, but does this PR include setting up redirects from the old location?

ShreyaThapa:
> @ShreyaThapa These resources will likely be moved again before the end of the year, but does this PR include setting up redirects from the old location?

Hey @spencerhunter, thanks for catching that! I hadn't added the redirects before. Just created a new commit with redirects to the current location of the pages.

spencerhunter (approved):

PR #545, opened 2023-09-26
Dev 1489: Add Connect docs, landing pages + new and existing pages

ShreyaThapa:


pages/docs/connect/webhook-events.mdx
```diff
@@ -0,0 +1,66 @@
...
+| created    | ISO-8601 timestamp of when the event was created.                                                                                                                                                                                                                                                                                                                                                                                                                                                |
+| topic      | Type of action that occurred within Dwolla. <br/> **Format:** `<resource>.<related_resource>:<type/action>` or `<resource>:<type/action>`. <br/> For example, the event topic for when a funding source belonging to an external party is removed is `external_party.funding_source:created`.                                                                                                                                                                                                    |
```
spencerhunter: @ShreyaThapa should we change this example to `account.transfer:created` since that's likely the most known?

ShreyaThapa: We can! My intention was to have an example that showed the casing using in the topic so developers can know what to expect. Perhaps we can add more examples like so? :

```suggestion
| topic      | Type of action that occurred within Dwolla. <br/> **Format:** `<resource>.<related_resource>:<type/action>` or `<resource>:<type/action>`. <br/> For example, the event topic for when a transfer is created is `account.transfer:created`. <br/> Other examples include: `account.funding_source:removed`, `external_party:created`.     
```

spencerhunter: Yeah, that sounds good. Small tweak to change the first colon ":" in the account.funding_source:removed topic name and it should be g2g

ShreyaThapa: Oops! Good catch! I edited it in the suggestion.

spencerhunter (approved):

PR #570, opened 2024-01-03
Global product selector

ShreyaThapa:
This branch includes:
- Addition of a new `SecondaryNavBar` component which is only displayed in a larger screen size
- Addition of a new global product selector drop-down in the `SideNav` (mobile experience) and the `SecondaryNavBar`
- Addition of a Category selector in the `SideNav` (mobile experience) which is used to as a replacement for the `SecondaryNavBar` on mobile screens
- Styles changes according to Figma Mocks
- Content migration placeholder/example; rest of the content is to be migrated to their new Product specific location within the `/pages` directory in a different branch `DEV-1499`


app/components/layout/SideNav.tsx
```diff
@@ -532,30 +549,12 @@ function SideNav({
...
-        // extracted from the active page's frontmatter. This ensures that the
-        // side nav reflects the selected product and its contents when navigating to a page
-        // from a URL.
-      }
-    }
   }, [pathname]);
 
+  useEffect(() => {
+    setSelectedSecondaryNavItem(selectedSecondaryNavItem);
+  }, [selectedSecondaryNavItem]);
```
Finholt: I believe this is redundant since the only way `selectedSecondaryNavItem` should change is if `setSelectedSecondaryNavItem` is called somewhere else

app/components/layout/Layout.tsx
```diff
@@ -137,6 +156,9 @@ export default function Layout({
   announcement?: JSX.Element;
 }) {
   const [sidebarToggled, setSidebarToggled] = useState(false);
+  const [selectedSecondaryNavItem, setSelectedSecondaryNavItem] = useState(
+    navItems[0]
```
Finholt: This default state is causing most pages to redirect back to `/docs/connect` when I navigate directly or refresh. Clearing the default value let me navigate more normally, but also made the dropdown in the sidenav on mobile not pre-populate with anything

pages/docs/balance/api-reference/index.mdx
Finholt, line 1: There's a link on `/docs/balance` that still links to `/api-reference` instead of to here

pages/docs/connect/api-reference/index.mdx
Finholt, line 1: There's a link on `/docs/connect` that still links to `/api-reference` instead of to here

app/components/layout/SecondaryNavBar.tsx
```diff
@@ -0,0 +1,142 @@
...
+                      color:
+                        router.asPath ===
+                        item.href(String(selectedProduct?.value))
+                          ? PURPLE_100
+                          : PURPLE_075,
+                      borderBottom:
+                        router.asPath ===
+                        item.href(String(selectedProduct?.value))
+                          ? `2px solid ${PURPLE_100}`
+                          : `2px solid transparent`,
```
Finholt, line 129: This only works if the path matches exactly, meaning child routes aren't representing the selected category properly. E.g. `/docs/connect/api-reference` correctly styles "API Reference" with the darker text and bottom border, but `/docs/connect/api-reference/accounts` does not

app/components/layout/SecondaryNavBar.tsx
```diff
@@ -0,0 +1,142 @@
...
+  setSelectedProduct,
+}: SecondaryNavBarProps) {
+  const router = useRouter();
+
+  useEffect(() => {
+    // Update the route and page based on the selectedSecondaryNavItem
+    if (selectedSecondaryNavItem && selectedProduct) {
+      router.push(selectedSecondaryNavItem.href(selectedProduct?.value));
+    }
+  }, [selectedSecondaryNavItem, selectedProduct]);
```
Finholt: After giving it some more thought, I think this approach that we landed on might be a bit backwards. Having the state of `selectedSecondaryNavItem` control the route means that if they ever get out of sync, the user will be redirected somewhere they don't necessarily want to go

For instance, if I click "Code Samples" in the secondary nav, click "Home" in the side nav, and then scroll down and click the "SDKS & Tools" link, I wind up back on the "Code Samples" page because that's still the value of `selectedSecondaryNavItem`

Rather than having the value of `selectedSecondaryNavItem` determine the value of the router, I think it would be safer to have the current path in the router determine the value of `selectedSecondaryNavItem`

Explained another way, here is the current flow:
1. User selects a category in either the secondary nav (on desktop) or the side nav (on mobile)
2. The value of `selectedSecondaryNavItem` is updated based on their selection
3. The selected value in the side nav dropdown is updated to reflect the new value of `selectedSecondaryNavItem`
4. This `useEffect` detects that the value of `selectedSecondaryNavItem` has changed and pushes a new route based on it

And this is the approach that I think might be better:
1. User selects a category in either the secondary nav (on desktop) or the side nav (on mobile)
2. A new route is pushed based on their selection
3. A new `useEffect` detects that the current path has changed and sets a new value for `selectedSecondaryNavItem` based on the new path
4. The selected value in the side nav dropdown is updated to reflect the new value of `selectedSecondaryNavItem`

That was a rather wordy explanation so if I failed to explain my thoughts clearly, please put something on my calendar and I can try to show you what I mean!

Finholt (approved):

PR #579, opened 2024-02-07
DEV-1534: Remove drop-ins from secondary nav based on Product Selection

ShreyaThapa:
- Adjusted the `navItems` array within `app/components/layout/Layout.tsx` based on the selected product, ensuring that dropIns is excluded from the array if the selectedProduct is "connect".
- Previously, the `id` in `SelectMui` was set using `useId`, but this led to potential duplicate `id` attributes when the component was re-rendered. To resolve this issue, we replaced the usage of `useId` with a `useRef` hook combined with a counter to generate unique id values for each instance of the `SelectMui` component.
- Updated the logic in the `_app.tsx` file to set the default selected product to "balance" if the user navigates directly to the "/docs/drop-in-components" path.


spencerhunter (approved):
🎉

PR #651, opened 2025-01-24
Dev 1666

ShreyaThapa:


spencerhunter:
@ShreyaThapa , Instead of moving this to the archive directory can we delete altogether? I want to make sure that this content doesn't get indexed for google search results, which would require removing it from the sitemap so crawlers don't pick it up.

spencerhunter (approved):

PR #660, opened 2025-06-16
remove info from VANs docs for updated implementation

spencerhunter:


pages/archive/virtual-account-numbers.mdx
```diff
@@ -155,114 +138,25 @@ Withdrawals from the Dwolla Balance will be represented by a new transfer where
...
-      "routingNumber": "string",             // of originating bank
-      "traceId": "string"
-    },
-    "destination": {
-      // same fields as source
-    }
-  }
-}
-```
+Additional details about an external transaction will be available in the when [retrieving the transfer](https://developers.dwolla.com/api-reference/transfers/retrieve) from the API. Either the source or the destination will be present in the object depending on the direction of the funds transfer.
```
ShreyaThapa:
```suggestion
Additional details about an external transaction will be available when [retrieving the transfer](https://developers.dwolla.com/api-reference/transfers/retrieve) from the API. Either the source or the destination will be present in the object depending on the direction of the funds transfer.
```

ShreyaThapa (changes requested):
Just a teensie 🤏 little change request!

ShreyaThapa (approved):

